Issue #7, opened 2021-05-07
Type system (common/base classes)


jjergus:
Flesh out the Hack classes that represent the GraphQL type system -- common type classes (e.g. scalar types like Int/String) and base classes (e.g. Object) that are subclassed by codegen.

This includes:

- how to deal with GraphQL's input vs. output types
  - scalars and enums are both
  - object/interface/union are output types
  - "input objects" (basically GraphQL's version of `dict`) are input types
- list type
  - for type safety, we may want to distinguish between lists of input types vs lists of output types
- nullability
  - design conundrum: GraphQL wraps nullable types to make them non-nullable (`Int` vs `Int!`) while Hack wraps non-nullable types to make them nullable (`int` vs `?int`)

This issue doesn't involve fully implementing everything, just fleshing out the basic structure. We should then open follow-up issues to implement things as needed.

PR #8, opened 2021-05-07
Initial support for handling inline arguments and variables

mwildehahn:
Addresses: https://github.com/mwildehahn/hack-graphql/issues/5

I skipped over validating the arguments just to get this working. I think we'll want to do something [similar to graphpinator](https://github.com/infinityloop-dev/graphpinator/blob/master/src/Graphpinator.php#L69) where they'll "normalize" the request, validating it and any specified arguments against the schema.

I looked at trying to just bring this logic in to get us started but it is a lot of PHP code that is at odds with our generated types.

**Testing**
- Added two test cases, one for variables and one for inline args.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,3 @@
 # hack-graphql
+
 Playground for a hack graphql server
--- a/src/BaseSchema.hack
+++ b/src/BaseSchema.hack
@@ -4,30 +4,54 @@ use namespace Facebook\TypeAssert;
 
 // TODO: this should be private
 abstract class BaseSchema {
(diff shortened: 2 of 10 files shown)


jjergus:
> I think we'll want to do something similar to graphpinator where they'll "normalize" the request, validating it and any specified arguments against the schema.

Yep this is part of the GraphQL spec: https://spec.graphql.org/draft/#sec-Validation

jjergus (approved):
Looks good to me. There are some more complicated cases that this won't handle (e.g. what if the value of an argument is a list that has a variable reference *inside* it) but I think my work on #7 should make those easier, so it's probably better to wait for that before dealing with any of the complicated cases.

PR #14, opened 2021-05-07
Flesh out some basic type system classes

### LINKED ISSUE #7: Type system (common/base classes)

jjergus:
Closes #7

Sorry for the somewhat long PR, but hopefully it should be easy to review the individual files, especially if you follow the order in which they appear in the class hierarchy (also there's probably more comments than real code here :) which should help):

- BaseType
  - InputType
    - NullableInputType
    - ListInputType
    - NamedInputType
      - IntInputType
      - StringInputType
  - OutputType
    - ListOutputType
    - NamedOutputType
      - IntOutputType
      - StringOutputType
      - ObjectType
        - (codegenned classes)

There are a few design choices here that are somewhat arbitrary (there are multiple options, each with some advantages and disadvantages). I chose options that roughly match our internal implementation, so my choices are based on some experience, but even if they were the correct choices for us, that doesn't necessarily mean that they are the best choices here, so feel free to disagree with anything!

- input vs output types
  - I put them into completely separate hierarchies, even though some input+output types are really the same type (e.g. IntInputType + IntOutputType) and could be represented by the same class implementing multiple interfaces, or something like that.
  - This is because input and output types end up being used in completely different ways in the GraphQL framework, and separating them like this allows us to make different design decisions for each hierarchy (e.g. input types have a generic type because it's very useful for type safety but output types don't; or how nullability is handled -- see below) instead of trying to make a single class that fits 2 very different uses.
- nullability (input types)
  - GraphQL defines nullable types to be the default and they can be "wrapped" to make them non-nullable (`Int` vs `Int!)`.
  - This is the exact opposite of Hack (`int` vs `?int`), so we have to choose one of the conflicting options.
  - Choosing the Hack model is AFAICT the only way to make input coercion type-safe.
- nullability (output types)
  - It might be natural to choose the same design for output types (for consistency), but I expect most fields to be nullable, so this would result in a lot of wrapping with no clear advantages.
  - Making output types nullable by default + a NonNullable wrapper could work pretty well, but it feels weird to choose a design that is literally the opposite of what we did for input types.
  - The third choice (used here) is to have a bool flag to indicate if the type is nullable, which is what I did for now.
  - This is the design choice that I am the *least* sure about. Probably worth revisiting after we start actually using the output types in more places.
- Every ["named type"](https://spec.graphql.org/draft/#sec-Wrapping-Types) is a singleton.
  - This has some advantages, e.g. types can be compared using `===`, might save some memory
  - But possibly also disadvantages, e.g. any type we instantiate stays in memory until the end of the request
  - It also gives us a neat API to get instances of types, e.g. `IntInputType::nullable()->nonNullableListOf()`


jjergus:
The output type classes here don't do anything yet, but the input type classes are sufficiently fleshed out so that we can implement argument handling (including the more complicated cases like lists with variables in them). I'll give that a try in a follow-up PR.

src/Types/Input/NamedInputType.hack
```diff
@@ -0,0 +1,42 @@
+namespace Slack\GraphQL\Types;
+
+/**
+ * Named type is any non-wrapping type.
+ *
+ * These are singletons. Get an instance using ::nullable() or ::nonNullable(), then call ->nullableListOf() or
+ * ->nonNullableListOf() to get a list, list of lists, etc. of this type.
+ *
+ * @see https://spec.graphql.org/draft/#sec-Wrapping-Types
```
mwildehahn, line 9: I love these links to the spec 🙏

src/Types/Input/Scalar/IntInputType.hack
```diff
@@ -0,0 +1,40 @@
...
+
+    const type TCoerced = int;
+    const string NAME = 'Int';
+
+    const int MIN_SAFE_VALUE = -2147483648;
+    const int MAX_SAFE_VALUE = 2147483647;
+
+    <<__Override>>
+    public function coerceValue(mixed $value): int {
+        // TODO: Should we allow coercion from string and/or float? e.g. something like:
```
mwildehahn, line 16: I like the idea of starting strict and then seeing if we need to do this. I prefer to be explicit but don't have a strong preference.

mwildehahn (approved):
Amazing! This looks great. The trade offs you mentioned all make sense.

src/Types/Input/Scalar/IntInputType.hack
jjergus: The argument here is that the value might come from a source that is not well typed, e.g. HTTP POST data -- so it could make sense to allow coercions that we wouldn't normally want to allow if everything was strict and well typed.

PR #18, opened 2021-05-08
Support mutations

mwildehahn:


tests/gen/Generated.hack
```diff
@@ -26,6 +27,20 @@ abstract final class Schema extends \Slack\GraphQL\BaseSchema {
 
     return await Dict\from_async($data);
   }
+
+  public static async function resolveMutation(
```
mwildehahn, line 31: I tried moving this to the base schema with a `resolveObject` method that both `resolveQuery` and `resolveMutation` called but got a type error since `self::resolveField` takes `self::THackType` as an argument.

src/Codegen/Generator.hack
```diff
@@ -119,6 +119,30 @@ class Query extends BaseObject<QueryField> {
     }
 }
 
+class Mutation extends BaseObject<MutationField> {
```
mwildehahn, line 122: This is pretty much the same as `Query`, we could pull this out and make it more DRY

ianhoffman:
It looks like this only supports mutations at the root level. Do we need to do anything else to support, say, mutating a user's name? I would think the `User` would need to do something like:

```hack
<<GraphQL\Object('User', 'User')>>
class User {
    <<GraphQL\MutationField('name', 'Set the user name')>>
    public function setName(string $name): string {
        $this->data['name'] = $name; // or whatever
    }
}
}
```

mwildehahn:
AFAIK mutations only live at the top level, similar to how we only expose a subset of top level fields you can query, ie:

```
type Query {
  user(id: ID!): User
}

type User {
   getFriends(): [List]
}
```

You can't get to `getFriends` without querying for a user first:

```
query { user(id: "1") { getFriends() { id } } }
```


To update a user we'd do something like:

```
type Mutation {
  updateUser(data: UpdateFields): User
}
```

jjergus:
Yep GraphQL fields are not meant to map directly to some database where they can be mutated field-by-field.

tests/gen/Generated.hack
jjergus: Might be possible with generics and `where` constraints, but I'm not sure.

PR #23, opened 2021-05-09
Support interfaces, v1

ianhoffman:
This is a pretty elementary first attempt at supporting interfaces. Basically, we just generate a new resolver for each interface, and then when generating a class which implements that interface we grab the interface fields and stick them in the class resolver as well. This doesn't produce the DRY-est code imaginable, but it's codegen so I'm not sure how much that matters.

I also have no idea how we'll support inline fragments - that can come later.

Any feedback would be much appreciated.


src/Codegen/Generator.hack
```diff
@@ -119,24 +119,28 @@ class Query extends BaseObject<QueryField> {
     }
 }
 
-class Object extends BaseObject<Field> {
+abstract class BaseClassishType<T as \Slack\GraphQL\__Private\Classish> extends BaseObject<Field> {
```
ianhoffman, line 122: There has to be a better name for this class 😓

src/Interface.hack
```diff
@@ -0,0 +1,3 @@
+namespace Slack\GraphQL;
+
+class GQLInterface extends \Slack\GraphQL\__Private\Classish {}
```
ianhoffman, line 3: The word `Interface` is reserved, sadly.

src/Codegen/Generator.hack
```diff
@@ -349,26 +357,34 @@ final class Generator {
...
 
+        $interfaces = dict[];
         $objects = vec[];
         $query_fields = vec[];
         foreach ($script->getDescendantsOfType(HHAST\ClassishDeclaration::class) as $class_decl) {
             if ($class_decl->hasAttribute()) {
                 $rc = new \ReflectionClass($class_decl->getName()->getText());
+
+                $graphql_interface = $rc->getAttributeClass(\Slack\GraphQL\GQLInterface::class);
+                if ($graphql_interface is nonnull) {
```
ianhoffman, line 368: Should we throw an error if the interface is actually a class? Or is it fine to do:

```hack
<<GQLInterface("Foo", "Foo")>>
abstract class Foo {
    <<Field("bar", "bar")>>
    public function bar(): string {
        return "bar";
    }
}
```

That seems generally useful - maybe we should require that GQLInterfaces be represented by Hack interfaces _or_ Hack abstract classes?

src/Codegen/Generator.hack
```diff
@@ -50,7 +50,7 @@ abstract class BaseObject<T as Field> implements GeneratableObjectType {
                 },
             )
             ->addDefault()
-            ->addLine("throw new \Error('Unknown field: '.\$field_name);")
+            ->addLine("throw new \Exception('Unknown field: '.\$field_name);")
```
ianhoffman, line 53: We can't catch an error using `expect(...)->toThrow()`, but we can catch an exception.

jjergus (approved):

src/Codegen/Generator.hack
```diff
@@ -349,26 +357,34 @@ final class Generator {
...
         $script = await HHAST\from_file_async(HHAST\File::fromPath($from_path));
 
+        $interfaces = dict[];
         $objects = vec[];
         $query_fields = vec[];
         foreach ($script->getDescendantsOfType(HHAST\ClassishDeclaration::class) as $class_decl) {
             if ($class_decl->hasAttribute()) {
                 $rc = new \ReflectionClass($class_decl->getName()->getText());
+
+                $graphql_interface = $rc->getAttributeClass(\Slack\GraphQL\GQLInterface::class);
```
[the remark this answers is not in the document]
jjergus: It seems reasonable to me that someone might create an interface that they were planning to start using later...

src/Codegen/Generator.hack
jjergus: The GraphQL spec calls objects, interfaces and unions "composite types".

src/Codegen/Generator.hack
jjergus: Oops we have a comment race condition :) I didn't see @mwildehahn's comment before commenting...I'm also OK with making things more restricted if it makes it easier for now, but I think we'll *eventually* want to support all these cases. 

The selling point of this framework is roughly "you write your Hack code naturally/take existing Hack code, then slap some <<GraphQL\... annotations on it and it magically works". If we add too many restrictions like "actually you can't put this on an abstract class, you must refactor it into an interface first" people can get frustrated about those. Such restrictions aren't completely unaviodable, but I think we want to minimize them over time.

src/Codegen/Generator.hack
ianhoffman: I like that. Will rename.

src/Interface.hack
ianhoffman: Will update to `InterfaceType` and rename `Object` to `ObjectType`.

ianhoffman:
> eaze of generating

This was generally "eaze of generating", lol.

src/Codegen/Generator.hack
jjergus: > Yea, that makes sense. Let's keep it like this and we can see how it feels to use. It's hard to have a strong opinion when we haven't used this yet smile

100% agreed. It's best to not overthink anything, because everything will most likely be rewritten at least once, probably multiple times before the final version.

PR #24, opened 2021-05-10
Use private namespace

### LINKED ISSUE #13: Move things into `__Private` that don't need to be public
- https://github.com/mwildehahn/hack-graphql/tree/main/src/Types
- https://github.com/mwildehahn/hack-graphql/tree/main/src/Codegen
- https://github.com/mwildehahn/hack-graphql/blob/main/src/BaseSchema.hack

should all be private

ianhoffman:
Closes #13 

Also:

* DRY out PlaygroundTest
* Move Playground.hack to the test dir


ianhoffman:
Gonna close this out and re-open when things settle down a bit. Too much movement right now to land this @mwildehahn

PR #29, opened 2021-05-10
Refactor execution, implement list + nullability

jjergus:
Sorry for another big one, but it's all intertwined so I found it hard to split into multiple parts. However, I did split this PR into several commits which can be reviewed individually, and if you review them in the order they're committed, I think it should be pretty straightforward.

- `ObjectType` now has `getFieldDefinition` instead of `resolveType` and `resolveField`. This makes it possible to resolve everything in a type-safe way without any runtime assertions!
- `OutputType` now handles nullability the same way as input types (instead of the bool flag), since that ended up the most natural way to implement everything in a type-safe way.
- The result of resolving any field is the new `FieldResult` class. This allows us to store the list of errors that need to be returned with the GraphQL response in addition to each field's value.
- The meat of this PR is the `resolveAsync` method on each output type class -- this is where all the important field-resolving logic lives.
- Generator needs cleaning up after this PR. I didn't do it because there are other PRs touching it and it would just cause more merge conflicts. I'll rebase this/clean it up after the other PRs are landed.


mwildehahn (approved):

src/Types/Output/EnumType.hack
```diff
@@ -0,0 +1,16 @@
...
+    abstract const \HH\enumname<this::THackType> HACK_ENUM;
+    const type TCoerced = string;
+
+    /**
+     * Enum values are surfaced to GraphQL clients as their name, not their internal value.
+     */
+    <<__Override>>
+    final protected function coerce(this::THackType $value): string {
+        $hack_enum = static::HACK_ENUM;
+        return $hack_enum::getNames()[$value];
```
ianhoffman, line 14: We catch UserFacingError in the superclass, but this will throw an InvalidIndex, right? Is that OK/desired?

src/Root.hack
```diff
@@ -0,0 +1,3 @@
+namespace Slack\GraphQL;
+
+final class Root {}
```
ianhoffman, line 3: +1

PR #33, opened 2021-05-11
Generate enum types

ianhoffman:
Supports generating enum types using a new `GraphQL\EnumType` attribute.


src/Codegen/Generator.hack
```diff
@@ -267,6 +272,31 @@ class QueryField extends Field {
...
 class MutationField extends QueryField {}
 
+class EnumType implements GeneratableObjectType {
+    public function __construct(
+        private DefinitionFinder\ScannedEnum $scanned_enum,
+        private \Slack\GraphQL\EnumType $enum_type,
+       ) {}
+
+    public function generateObjectType(HackCodegenFactory $cg): CodegenClass {
+        return $cg->codegenClass($this->enum_type->getType())
```
ianhoffman, line 282: This is quite similar to `generateObjectType` for `CompositeType`; maybe we can dry this up later.

src/Codegen/Generator.hack
```diff
@@ -267,6 +277,31 @@ class QueryField extends Field {
...
 class MutationField extends QueryField {}
 
+class EnumOutputType implements GeneratableObjectType {
+    public function __construct(
+        private DefinitionFinder\ScannedEnum $scanned_enum,
+        private \Slack\GraphQL\EnumType $enum_type,
+       ) {}
+
+    public function generateObjectType(HackCodegenFactory $cg): CodegenClass {
+        return $cg->codegenClass($this->enum_type->getOutputType())
```
ianhoffman, line 287: We should start DRYing up all these different generators 😓

ianhoffman:
Hmm, it looks like I should have rebased onto main instead of merging it in. 😓  Looking into fixing this.

mwildehahn (approved):
LGTM!

src/Codegen/Generator.hack
ianhoffman: Definitely

PR #39, opened 2021-05-12
non-nullable fields + errors

jjergus:
Adds codegen support for non-nullable fields (using the new attribute `KillsParentOnException` -- see its docblock for explanation) and a bunch of test cases. The test cases are actually the important part here (I wanted to ensure that error handling is implemented correctly), the support for non-nullable fields is secondary (it was impossible to write the tests without it) and I wouldn't necessarily recommend using them.

That said, non-nullable fields *do* have advantages (mainly, clients don't need as many null checks everywhere, which client developers like a lot), so maybe we shouldn't discourage them as much.

At FB, we heavily discourage non-nullable fields, but that is mostly for historical reasons and not necessarily the right call everywhere.

If you decide that non-nullable fields are fine, I'd recommend renaming `KillsParentOnException` to something less scary, possibly just `NonNullable`.

### DIFF
--- a/src/Codegen/Generator.hack
+++ b/src/Codegen/Generator.hack
@@ -144,7 +144,10 @@ class Field {
         $hb->addLine('return new GraphQL\\FieldDefinition(')->indent();
 
         // Field return type
-        $type_info = output_type($this->reflection_method->getReturnTypeText());
+        $type_info = output_type(
+            $this->reflection_method->getReturnTypeText(),
+            $this->reflection_method->getAttributeClass(\Slack\GraphQL\KillsParentOnException::class) is nonnull,
+        );
         $hb->addLinef('%s,', $type_info['type']);
 
         // Arguments
--- a/src/Field.hack
+++ b/src/Field.hack
@@ -7,3 +7,26 @@ class Field implements \HH\MethodAttribute {
         return $this->name;
     }
 }
+
+/**
+ * Must be used in combination with GraphQL\Field, on a method with a non-nullable return type in Hack. If present, the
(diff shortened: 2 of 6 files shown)


src/Codegen/Generator.hack
```diff
@@ -144,7 +144,10 @@ class Field {
         $hb->addLine('return new GraphQL\\FieldDefinition(')->indent();
 
         // Field return type
-        $type_info = output_type($this->reflection_method->getReturnTypeText());
+        $type_info = output_type(
+            $this->reflection_method->getReturnTypeText(),
+            $this->reflection_method->getAttributeClass(\Slack\GraphQL\KillsParentOnException::class) is nonnull,
```
mwildehahn, line 149: lol this name 🤣

src/Field.hack
```diff
@@ -7,3 +7,26 @@ class Field implements \HH\MethodAttribute {
         return $this->name;
     }
 }
+
+/**
+ * Must be used in combination with GraphQL\Field, on a method with a non-nullable return type in Hack. If present, the
```
mwildehahn, line 12: i love these descriptions

mwildehahn (approved):

src/Codegen/Generator.hack
jjergus: This is actually the name we use internally, but like I said above it might be a good idea to change it to something less scary -- non-nullable fields are much less dangerous if they are used properly from the beginning. We had to discourage them mostly because we support lots of old clients with old assumptions.

I opened #42 for possible follow-up

PR #40, opened 2021-05-12
Support Enum Input Types

ianhoffman:


src/Resolver.hack
```diff
@@ -69,14 +69,14 @@ final class Resolver {
         $schema = $this->schema;
 
         if ($operation_name is nonnull) {
-            assert(
+            invariant(
```
ianhoffman, line 72: `assert` must be post HHVM 4.56

tests/ErrorTest.hack
```diff
@@ -244,6 +244,38 @@ final class ErrorTest extends PlaygroundTest {
...
+            'invalid enum values with arguments' => tuple(
+                'query { takes_favorite_color(favorite_color: $favorite_color) }',
+                dict['favorite_color' => 'foo'],
+                shape(
+                    'data' => dict[
+                        'takes_favorite_color' => null
+                    ],
+                    'errors' => vec[
+                        shape(
+                            'message' => 'Caught exception while resolving field.',
```
ianhoffman, line 273: This error is kind of lame — would it make more sense to call `coerceValue()` instead of `assertValidVariableValue()` in `InputType::coerceNode()`, @jjergus ? The methods do nearly the same thing in all cases, except `coerceValue()` is a little more permissive and has better error messages.

tests/gen/Generated.hack
```diff
@@ -150,6 +157,56 @@ final class User extends \Slack\GraphQL\Types\ObjectType {
   }
 }
 
+final class OutputTypeTest extends \Slack\GraphQL\Types\ObjectType {
```
ianhoffman, line 160: Weird that sort order seems non-deterministic for DefinitionFinder. We should sort our objects before generating code.

src/Types/Input/EnumInputType.hack
```diff
@@ -0,0 +1,33 @@
...
+
+abstract class EnumInputType extends NamedInputType {
+
+    <<__Enforceable>>
+    abstract const type TCoerced as arraykey;
+    abstract const \HH\enumname<this::TCoerced> HACK_ENUM;
+
+    <<__Override>>
+    public function coerceValue(mixed $value): this::TCoerced {
+        if (!$value is this::TCoerced) {
```
mwildehahn, line 15: Do `is` and `as` replace `coerce` and `assert` from enums? I thought you'd need to use those but it looks like this works too.

tests/gen/Generated.hack
mwildehahn: Yea, I write out the top level objects first but would be good to sort this too. Add an issue!

src/Codegen/Types.hack
```diff
@@ -75,6 +80,20 @@ function output_type(
     return shape('type' => Str\strip_prefix($class, 'Slack\\GraphQL\\').$suffix, 'needs_await' => $needs_await);
 }
 
+/**
+ * Get the input class for a hack type which is not a primitive.
+ */
+function get_input_class(string $hack_type): ?string {
+    $rc = new \ReflectionClass($hack_type);
```
mwildehahn, line 87: For https://github.com/mwildehahn/hack-graphql/pull/41/files I ended threaded through the inputs we collected but I like this more since it means we don't have to pass through a bunch of dicts here. The one thing is that we'll need to handle catching the error raised if this isn't a class, ie. I think this will look something like:

```
function get_input_type(string $hack_type): ?string {
  $rc = null;
  try {
    $rc = new \ReflectionClass($hack_type);
  catch (WhateverReflectionException) {}
  
  if ($rc is nonnull) {...}
  
  $rt = null;
  try {
    $rt = new \ReflectionType($hack_type);
  catch (WhateverReflectionException) {}
  
  if ($rt is nonnull) {...}
  
  ...
}
```

I can adjust this in my PR after you merge this

mwildehahn (approved):
LGTM!

tests/gen/Generated.hack
mwildehahn: Yea, I write out the top level objects first but would be good to sort this too. Add an issue!

src/Resolver.hack
jjergus: This is [GraphQL\assert](https://github.com/mwildehahn/hack-graphql/blob/main/src/UserFacingError.hack#L5) which raises a `UserFacingError`.

Though I remember that there used to be a built-in `assert` in HHVM so maybe this conflicts with it and we need to rename it :(

tests/ErrorTest.hack
jjergus: Indeed, if this error shows up anywhere besides its intended use (i.e. when the actual Hack method that is annotated with `<<GraphQL\Field>>` throws an exception), that is a bug.

In this case, the bug is that we have not yet implemented [this TODO](https://github.com/mwildehahn/hack-graphql/blob/main/src/Resolver.hack#L68) and query validation. After that is implemented, `assertValidVariableValue()` should be guaranteed to never throw (there's also a comment about it [here](https://github.com/mwildehahn/hack-graphql/blob/main/src/Types/Input/InputType.hack#L41)).

src/Types/Input/EnumInputType.hack
jjergus: I think we should be expecting the enum const name here, not the internal value (which is what this check would accept). Probably something like

```Hack
$enum = static::HACK_ENUM;
GraphQL\assert(
  C\contains_key($enum::getValues(), $value),
  'Expected ...'
);
return $enum::getValues()[$value];
```

src/Types/Input/EnumInputType.hack
```diff
@@ -0,0 +1,33 @@
...
+        return $value;
+    }
+
+    <<__Override>>
+    final public function coerceNonVariableNode(
+        Value\Value $node,
+        dict<string, mixed> $variable_values,
+    ): this::TCoerced {
+        $value = $node->getRawValue();
+        if (!$node is Value\EnumLiteral || !$value is this::TCoerced) {
```
jjergus, line 27: Same here.

The additional check (`!$node is Value\EnumLiteral`) is correct, we should indeed be rejecting e.g. a string literal even if its value would be a valid enum const name.

```GraphQL
some_field(enum_arg: FOO) # good
some_field(enum_arg: "FOO") # bad
```

(probably worth adding to the test)

tests/BasicTest.hack
```diff
@@ -57,6 +57,16 @@ final class BasicTest extends PlaygroundTest {
                 dict[],
                 dict['pokeUser' => dict['id' => 2]],
             ),
+            'enum arguments' => tuple(
+                'query { takes_favorite_color(favorite_color: red) }',
```
jjergus, line 61: This should accept `RED` but not `red`. Maybe we could change the `FavoriteColor` enum to be an int enum to make this distinction clearer:

```Hack
enum FavoriteColor: int {
  RED = 1;
  BLUE = 2;
}
```

This also makes it much more obvious why accepting the const names is better than using the internal values; we wouldn't want GraphQL clients to have to write `favorite_color: 2`

jjergus (approved):

src/Resolver.hack
ianhoffman: Ah, gotcha - will fix.

src/Types/Input/EnumInputType.hack
ianhoffman: Good point about `contains_key`, will update.

src/Codegen/Types.hack
ianhoffman: Updating.

tests/ErrorTest.hack
ianhoffman: Ah, makes sense - I'll leave this as-is for now, then.

src/Codegen/Types.hack
```diff
@@ -75,6 +80,21 @@ function output_type(
...
+ * Get the input class for a hack type which is not a primitive.
+ */
+function get_input_class(string $hack_type): ?string {
+    try {
+        $rc = new \ReflectionClass($hack_type);
+        $graphql_enum = $rc->getAttributeClass(\Slack\GraphQL\EnumType::class);
+        if ($graphql_enum) {
+            return $graphql_enum->getInputType();
+        }
+    } catch (\ReflectionException $_e) {}
```
ianhoffman, line 93: CC @mwildehahn

src/Types/Input/EnumInputType.hack
```diff
@@ -0,0 +1,47 @@
...
+    <<__Override>>
+    public function coerceValue(mixed $value): this::TCoerced {
+        $enum = static::HACK_ENUM;
+        GraphQL\assert(
+            $value is string && C\contains_key($enum::getValues(), $value),
+            'Expected a valid value for %s, got %s',
+            static::NAME,
+            (string)$value
+        );
+        return $enum::getValues()[$value as string];
```
jjergus, line 23: nitpick: We can avoid this `as` (I try to avoid runtime assertion as they can be a dangerous way to bypass the typechecker) if you use `if (...) throw new UserFacingException(...)` instead of `GraphQL\assert`

This runtime assertion is clearly fine though, so I don't feel too strongly about it.

jjergus (approved):
You don't have to request review again if everyone already approved it :)

src/Types/Input/EnumInputType.hack
ianhoffman: Will update!

ianhoffman:
> You don't have to request review again if everyone already approved it :)

Fair - just wanted to ensure the changes are approved too.

PR #41, opened 2021-05-12
Add support for `GraphQL\InputType` attribute

mwildehahn:
Allows tagging of shapes with `<<GraphQL\InputType>>` to support accepting those shapes as input. GraphQL spec: https://graphql.org/learn/schema/#input-types.

> The fields on an input object type can themselves refer to input object types, but you can't mix input and output types in your schema. Input object types also can't have arguments on their fields.

I didn't add any checks for this yet.

One other note, this requires type assertion to verify the incoming dict coerces to the defined type. Not sure if there is a better way to do this.

### DIFF
--- a/src/InputType.hack
+++ b/src/InputType.hack
@@ -6,7 +6,7 @@ namespace Slack\GraphQL;
 * Annotate shapes with this attribute to support accepting them as inputs within
 * mutations.
 */
-class InputType implements \HH\TypeAliasAttribute {
+class InputObjectType implements \HH\TypeAliasAttribute {
     public function __construct(private string $type, private string $description) {}
 
     public function getType(): string {
--- a/src/InputType.hack
+++ b/src/InputType.hack
@@ -1,0 +1,15 @@
+namespace Slack\GraphQL;
+
+/**
+* GraphQL input type: https://graphql.org/learn/schema/#input-types
+*
+* Annotate shapes with this attribute to support accepting them as inputs within
+* mutations.
+*/
+class InputType implements \HH\TypeAliasAttribute {
+    public function __construct(private string $type, private string $description) {}
+
+    public function getType(): string {
+        return $this->type;
+    }
+}
--- a/src/Resolver.hack
+++ b/src/Resolver.hack
@@ -78,7 +78,7 @@ final class Resolver {
         } else {
             \Slack\GraphQL\assert(
                 C\count($request->getOperations()) === 1,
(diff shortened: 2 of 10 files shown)


mwildehahn:
I'm going to refactor this a bit after https://github.com/mwildehahn/hack-graphql/pull/40 gets merged so we can do this: https://github.com/mwildehahn/hack-graphql/pull/40#discussion_r631171213 and not have to collect all inputs and thread them through.

src/InputType.hack
```diff
@@ -0,0 +1,15 @@
+namespace Slack\GraphQL;
+
+/**
+* GraphQL input type: https://graphql.org/learn/schema/#input-types
+*
+* Annotate shapes with this attribute to support accepting them as inputs within
+* mutations.
+*/
+class InputType implements \HH\TypeAliasAttribute {
```
jjergus, line 9: naming nitpick: "input type" can mean any of scalars, enums, input objects

Maybe "InputObjectType" or just "InputObject"?

The usage of "object" here can be a bit confusing for people who don't work in JavaScript; in JS "object" is basically a map/dict.

src/Resolver.hack
```diff
@@ -69,16 +69,16 @@ final class Resolver {
         $schema = $this->schema;
 
         if ($operation_name is nonnull) {
-            assert(
+            invariant(
```
jjergus, line 72: See my comment in #40, it's important for this to throw a `UserFacingError`

src/Types/Input/InputObjectType.hack
```diff
@@ -0,0 +1,37 @@
...
+    final public function getName(): string {
+        return static::NAME;
+    }
+
+    <<__Override>>
+    final public function coerceNonVariableNode(
+        \Graphpinator\Parser\Value\Value $node,
+        dict<string, mixed> $variable_values,
+    ): TCoerced {
+        return $this->coerceValue($node->getRawValue());
```
jjergus, line 22: I think this won't work if there are variable references nested inside the input object, but we can fix that later. I have an idea for what might be a good implementation so I could give it a try after you merge this.

```GraphQL
mutation CreateUser($name: String!) {
    createUser(input: {name: $name}) {
        id
        is_active
        name
        team {
            name
        }
    }
}
```

jjergus (approved):

src/InputType.hack
mwildehahn: I'll go with InputObjectType for now to stay consistent

PR #51, opened 2021-05-12
input object coercion

jjergus:
- supports variables nested inside an input object literal (see new test cases)
- supports enum fields (which need to be coerced between const name <-> internal value, which previously wouldn't happen)
- the generated code is type-safe without runtime assertions (TypeAssert), so if we ever introduce a bug into the codegen, Hack should catch it

This doesn't yet handle optional/nullable fields, I'll do those separately tomorrow.

I moved InputObjectType out of Generator.hack since that file was getting very long.

The easiest way to review this is probably look at the changes to the codegen output (Generated.hack) first, the rest should become pretty obvious after that.

### DIFF
--- a/src/Codegen/InputObjectType.hack
+++ b/src/Codegen/InputObjectType.hack
@@ -1,0 +1,98 @@
+namespace Slack\GraphQL\Codegen;
+
+use namespace HH\Lib\{Keyset, Str};
+use type Facebook\HackCodegen\{CodegenClass, CodegenMethod, HackBuilderValues, HackCodegenFactory};
+
+class InputObjectType implements GeneratableClass {
+    public function __construct(
+        private \ReflectionTypeAlias $reflection_type_alias,
+        private \Slack\GraphQL\InputObjectType $input_type,
+    ) {}
+
+    public function getType(): string {
+        return $this->input_type->getType();
+    }
+
+    public function generateClass(HackCodegenFactory $cg): CodegenClass {
+        $hack_type = $this->reflection_type_alias->getName();
+
+        $class = $cg->codegenClass($this->input_type->getType())
+            ->setIsFinal()
+            ->setExtendsf('\%s', \Slack\GraphQL\Types\InputObjectType::class)
+            ->addTypeConstant(
+                $cg->codegenTypeConstant('TCoerced')->setValue('\\'.$hack_type, HackBuilderValues::literal()),
+            );
+
+        $name_const = $cg
+            ->codegenClassConstant('NAME')
+            ->setValue($this->input_type->getType(), HackBuilderValues::export());
+
+        $class->addConstant($name_const);
+
+        $ts = $this->reflection_type_alias->getResolvedTypeStructure();
+        invariant(
+            $ts['kind'] === \HH\TypeStructureKind::OF_SHAPE && !idx($ts, 'nullable', false),
+            'Input objects can only be generated from type aliases of a non-nullable shape type, got %s.',
+            TypeStructureKind::getNames()[$ts['kind']],
+        );
+
+        $class->addConstant(
+            $cg->codegenClassConstant('keyset<string> FIELD_NAMES')
+                ->setValue(Keyset\keys($ts['fields']), HackBuilderValues::export()),
+        );
+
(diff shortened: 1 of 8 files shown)


src/Codegen/InputObjectType.hack
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        switch ($ts['kind']) {
+            case TypeStructureKind::OF_INT:
+                return 'HH\int';
+            case TypeStructureKind::OF_STRING:
+                return 'HH\string';
+            case TypeStructureKind::OF_BOOL:
+                return 'HH\bool';
+            case TypeStructureKind::OF_VEC:
```
ianhoffman, line 86: Do we need another case here to handle nested shapes?

src/Types/Input/EnumInputType.hack
```diff
@@ -25,6 +25,8 @@ abstract class EnumInputType extends NamedInputType {
     }
 
     final protected function assertValidVariableValue(mixed $value): this::TCoerced {
+        // TODO: This should just be $value as this::TCoerced, once variables are properly coerced at the beginning
+        // of execution. Otherwise we'd try (and fail) to coerce an already-coerced value here.
```
ianhoffman, line 29: Good call out!

src/Codegen/InputObjectType.hack
ianhoffman: Ah, I guess `$alias` takes care of that.

ianhoffman (approved):

src/Codegen/InputObjectType.hack
```diff
@@ -0,0 +1,98 @@
+namespace Slack\GraphQL\Codegen;
+
+use namespace HH\Lib\{Keyset, Str};
+use type Facebook\HackCodegen\{CodegenClass, CodegenMethod, HackBuilderValues, HackCodegenFactory};
+
+class InputObjectType implements GeneratableClass {
```
mwildehahn, line 6: Nice, thanks for pulling this out

src/Codegen/InputObjectType.hack
mwildehahn, line 86: Should we handle keyset and just treat that as vec here?

tests/InputTypeTest.hack
```diff
@@ -94,6 +94,95 @@ final class InputTypeTest extends PlaygroundTest {
...
+                        'id' => 3,
+                        'is_active' => false,
+                        'name' => 'New User',
+                        'team' => dict['name' => 'New Team'],
+                    ],
+                ],
+            ),
+
+            'input object literal with variable references' => tuple(
+                'mutation CreateUser(
```
mwildehahn, line 124: I can't wait to use this 😄

mwildehahn (approved):
🤯

src/Codegen/InputObjectType.hack
jjergus: Yeah the TypeStructure API is very confusing :( Various types get surfaced in very different ways. I'm honestly not sure if these cases cover everything, but one of the test input objects has a nested input objects so that part works...probably.

jjergus: This could actually get quite tricky. If we support keyset fields on input object shapes, there would no longer be 1:1 mapping between GraphQL types and Hack types (now `[Int]` could be either `vec<int>` or `keyset<int>`), which can complicate the coercion logic significantly, especially when variables are involved -- e.g. a GraphQL variable `$foo: [Int]` could be used in multiple places in the GraphQL query, one of which needs coercion to `vec` and the other to `keyset`. This means we could no longer coerce variables at the beginning of execution and then just use them, there would have to be an extra coercion step at each usage...

jjergus: It is definitely doable though, if people think it would be useful.

jjergus: Also, this would actually be pretty easy for *output* types. We can support `<<GraphQL\Field>>` on methods that return keyset/Container/... easily.

mwildehahn: Yea, I think the output types make more sense to me since it's fine to represent that as a vec in GraphQL

PR #52, opened 2021-05-13
Improve interface codegen

ianhoffman:
This PR allows codegen-ing interfaces which extend interfaces. We now traverse the entire object graph to resolve fields for each class or interface.


src/Codegen/FieldResolver.hack
```diff
@@ -0,0 +1,73 @@
...
+        $parents = $class->getInterfaceNames();
+        $parent_class_name = $class->getParentClassName();
+        if ($parent_class_name is nonnull) {
+            $parents[] = $parent_class_name;
+        }
+
+        foreach ($parents as $parent) {
+            $parent_class = $this->scanned_classes[$parent] ?? null;
+            if ($parent_class) {
+                $fields = Dict\merge($fields, $this->resolveClass($parent_class));
```
jjergus, line 50: At some point we'll probably want some extra validation here, instead of just blindly merging the dicts. For example:

- a method with the same name should not have `<<GraphQL\Field(...)>>` with different field names across the hierarchy
- arguments should have the same names (since GraphQL has named arguments, unlike Hack)

A simpler option could be to just enforce that `<<GraphQL\Field(...)>>` can only be used once for each method name across the hierarchy -- e.g. if there's `<<GraphQL\Field('foo')>>` on an interface, adding `<<GraphQL\Field>>` on the overridden method in the subclass would be illegal.

jjergus (approved):

src/Codegen/FieldResolver.hack
```diff
@@ -0,0 +1,73 @@
...
+
+    private function resolveClass(DefinitionFinder\ScannedClassish $class): dict<string, Field> {
+        if (C\contains_key($this->resolved_fields, $class->getName())) {
+            return $this->resolved_fields[$class->getName()];
+        }
+
+        $fields = dict[];
+
+        $parents = $class->getInterfaceNames();
+        $parent_class_name = $class->getParentClassName();
```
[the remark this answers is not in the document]
ianhoffman: Will handle traits separately in a follow-up.

src/Codegen/FieldResolver.hack
ianhoffman: > A simpler option could be to just enforce that <<GraphQL\Field(...)>> can only be used once for each method name across the hierarchy -- e.g. if there's <<GraphQL\Field('foo')>> on an interface, adding <<GraphQL\Field>> on the overridden method in the subclass would be illegal.

This seems effectively the same as what you detailed above, and therefore preferable. I'll file an issue.

src/playground/InterfaceTestObjects.hack
```diff
@@ -0,0 +1,48 @@
...
+    <<GraphQL\QueryRootField('getInterfaceB', 'Root field to get an instance of InterfaceB')>>
+    public static function getInterfaceB(): InterfaceB {
+        return new self();
+    }
+
+    public function foo(): string {
+        return 'foo';
+    }
+
+    <<GraphQL\Field('bar', 'bar')>>
```
jjergus, line 39: This is an example where a field from an interface is overridden. The two options I suggested are:

- validate that this is consistent with the overridden declaration
- make this illegal (simpler option, but will reject perfectly reasonable declarations like this one)

ianhoffman: Totally. I filed an issue to follow up on this here: https://github.com/mwildehahn/hack-graphql/issues/59

Issue #53, opened 2021-05-13
Support annotating shapes with GraphQL\ObjectType


mwildehahn:

ianhoffman:
Just a question here - how do you imagine fields working? Is each member of the shape a field (I assume the answer is yes). And then do these fields have no descriptions?

jjergus:
Yeah each shape field would be a GraphQL field. This will add the restriction that each shape field must be of a type that can be translated to a valid GraphQL input **and** output type (scalars, enums, other shapes that are also annotated with both ObjectType and InputObjectType; but no input objects, or object/interface output types) but that should be fine.

> And then do these fields have no descriptions?

For things that can't have attributes (shape fields, but also e.g. method arguments) a decent option is to use docblocks/comments for descriptions (parsed via HHAST probably).

jjergus:
Alternatively something like

```Hack
<<
  GraphQL\Object('User'),
  GraphQL\InputObject<'UserInput'>,
  GraphQL\Descriptions(dict[
    'name' => 'Full name of the user',
    // ...
  ])
>>
newtype User = shape(
  'name' => string,
  // ...
);

ianhoffman:
Closing this and moving the descriptions stuff to #64 .

PR #61, opened 2021-05-13
Support annotating shapes with GraphQL\ObjectType

ianhoffman:
Implements much of #53 .

We don't yet support descriptions because we don't yet have a use for them. They'll be used by the introspection system at some point (hopefully soon), and we can revisit then.

The generator code is getting increasingly gnarly. This new `OutputObjectType.hack` `GeneratableClass` looks a lot like both the code for `InputObjectType` and `CompositeObject`. We should find some nice way to share code between these types, but I'm punting for now.

### DIFF
--- a/src/Types/Output/NamedOutputType.hack
+++ b/src/Types/Output/NamedOutputType.hack
@@ -11,7 +11,6 @@ namespace Slack\GraphQL\Types;
 <<__ConsistentConstruct>>
 abstract class NamedOutputType extends OutputType<this::THackType, this::TCoerced> {
 
-    <<__Enforceable>>
     abstract const type THackType;
     abstract const type TCoerced;
     abstract const string NAME;
--- a/src/Codegen/Generator.hack
+++ b/src/Codegen/Generator.hack
@@ -437,9 +437,14 @@ final class Generator {
         foreach ($input_types as $type) {
             $rt = new \ReflectionTypeAlias($type->getName());
             $graphql_input = $rt->getAttributeClass(\Slack\GraphQL\InputObjectType::class);
-            if ($graphql_input is null) continue;
+            if ($graphql_input is nonnull) {
+                $inputs[] = new InputObjectType($rt, $graphql_input);
+            }
 
-            $inputs[] = new InputObjectType($rt, $graphql_input);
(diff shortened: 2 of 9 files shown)


src/Types/Output/NamedOutputType.hack
```diff
@@ -11,7 +11,6 @@ namespace Slack\GraphQL\Types;
 <<__ConsistentConstruct>>
 abstract class NamedOutputType extends OutputType<this::THackType, this::TCoerced> {
 
-    <<__Enforceable>>
     abstract const type THackType;
```
ianhoffman, line 14: Getting rid of the `__Enforceable` attribute here both didn't seem to break anything and allowed me to repurposes `ObjectType` for objects generated from shapes even if they contain reified parameters.

mwildehahn (approved):

src/Types/Output/NamedOutputType.hack
jjergus: Good catch, this was only needed before #29

jjergus (approved):

jjergus:
> We should find some nice way to share code between these types, but I'm punting for now.

Yep I would definitely recommend doing this soon, probably the sooner it's done the easier it is. It doesn't have to be perfect, I imagine there will be another refactor needed after some time as new requirements are added...

Good news is that the codegen should be easy to refactor, because it can be tested very easily (if it generates the same output it's probably correct).

ianhoffman:
> > We should find some nice way to share code between these types, but I'm punting for now.
> 
> Yep I would definitely recommend doing this soon, probably the sooner it's done the easier it is. It doesn't have to be perfect, I imagine there will be another refactor needed after some time as new requirements are added...
> 
> Good news is that the codegen should be easy to refactor, because it can be tested very easily (if it generates the same output it's probably correct).

True. I'll see if I can't find some time for that this weekend.

jjergus:
It's not that urgent, definitely don't spend a weekend working on it :)

ianhoffman:
> It's not that urgent, definitely don't spend a weekend working on it :)

Ha, I have tomorrow off and most people I know don't, so I expect I'll be able to find a couple hours for this. But I'll make sure not to work too hard :)

PR #62, opened 2021-05-13
codegen dicts of input/output types

jjergus:
...so we can use them for introspection, validation, variable coercion, etc.

I made the generics on InputType and OutputType covariant (`+T`) or contravariant (`-T`) as needed. This is so that we can refer to e.g. a list of arbitrary input types as `InputType<mixed>`, which is impossible if the generic type is invariant.

### DIFF
--- a/src/BaseSchema.hack
+++ b/src/BaseSchema.hack
@@ -3,6 +3,8 @@ namespace Slack\GraphQL;
 // TODO: this should be private
 abstract class BaseSchema {
     const SUPPORTS_MUTATIONS = false;
+    abstract const dict<string, classname<Types\NamedInputType>> INPUT_TYPES;
+    abstract const dict<string, classname<Types\NamedOutputType>> OUTPUT_TYPES;
 
     abstract public static function resolveQuery(
         \Graphpinator\Parser\Operation\Operation $operation,
--- a/src/Codegen/Generator.hack
+++ b/src/Codegen/Generator.hack
@@ -3,7 +3,7 @@ namespace Slack\GraphQL\Codegen;
 use namespace Slack\GraphQL\Types;
 use namespace Facebook\HHAST;
 use namespace Facebook\DefinitionFinder;
-use namespace HH\Lib\{Str, Vec, C};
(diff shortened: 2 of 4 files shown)


ianhoffman:
> I made the generics on InputType and OutputType covariant (+T) or contravariant (-T) as needed. This is so that we can refer to e.g. a list of arbitrary input types as InputType<mixed>, which is impossible if the generic type is invariant.

After reading this PR I understand co- and contravariance more than I ever have before, which is to say, I 50% understand them.

ianhoffman (approved):

jjergus:
> After reading this PR I understand co- and contravariance more than I ever have before, which is to say, I 50% understand them.

Heh, I just realized that part of this PR was actually unnecessary, since the lists should only contain NamedInputType/NamedOutputType, and those don't have any generics. I'll revert that part to keep things simple.

ianhoffman:
> > After reading this PR I understand co- and contravariance more than I ever have before, which is to say, I 50% understand them.
> 
> Heh, I just realized that part of this PR was actually unnecessary, since the lists should only contain NamedInputType/NamedOutputType, and those don't have any generics. I'll revert that part to keep things simple.

Oh, true! That is simpler.

jjergus:
> After reading this PR I understand co- and contravariance more than I ever have before, which is to say, I 50% understand them.

If you wanna go from 50% to maybe 80%, https://docs.hhvm.com/hack/generics/variance has a decent overview, but I guarantee no one will ever 100% understand this :)

ianhoffman:
> If you wanna go from 50% to maybe 80%, https://docs.hhvm.com/hack/generics/variance has a decent overview, but I guarantee no one will ever 100% understand this :)

Ha, I’ve read it a few times over the past year and it has helped somewhat, but I think encountering real-world use-cases is illuminating. It’s slowly coming together 😅

ianhoffman:
The array of animals example is pretty helpful: a read-only array of cats can be passed to a function which expects an array of animals, and a write-only array of animals can be passed to a function which expects an array of cars.

PR #63, opened 2021-05-14
coerce variable values before execution

### LINKED ISSUE #44: coerce variables at the beginning of execution
https://github.com/mwildehahn/hack-graphql/blob/main/src/Resolver.hack#L68

This is required for `assertValidVariableValue()` to be correct: https://github.com/mwildehahn/hack-graphql/blob/main/src/Types/Input/InputType.hack#L40

jjergus:
This should get rid of most of the spurious errors from `assertValidVariableValue()`. They may still happen though (until we have query validation), if a variable is used as the value of a field argument of an incompatible type.

Closes #44

### DIFF
--- a/src/Types/Input/EnumInputType.hack
+++ b/src/Types/Input/EnumInputType.hack
@@ -24,13 +24,6 @@ abstract class EnumInputType extends NamedInputType {
         return $enum::getValues()[$value];
     }
 
-    final protected function assertValidVariableValue(mixed $value): this::TCoerced {
-        // TODO: This should just be $value as this::TCoerced, once variables are properly coerced at the beginning
-        // of execution. Otherwise we'd try (and fail) to coerce an already-coerced value here.
-        $enum = static::HACK_ENUM;
-        return $enum::getValues()[$value as string];
-    }
-
     <<__Override>>
     final public function coerceNonVariableNode(
         Value\Value $node,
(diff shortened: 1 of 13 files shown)


src/Types/Input/EnumInputType.hack
```diff
@@ -25,10 +25,7 @@ abstract class EnumInputType extends NamedInputType {
     }
 
     final protected function assertValidVariableValue(mixed $value): this::TCoerced {
```
ianhoffman, line 27: You can actually delete this entirely and make the parent class's implementation final, I think.

src/Types/Input/InputType.hack
```diff
@@ -2,15 +2,21 @@ namespace Slack\GraphQL\Types;
...
  *
  * @see https://spec.graphql.org/draft/#sec-Input-and-Output-Types
  */
-abstract class InputType<TCoerced> extends BaseType {
+abstract class InputType<+TCoerced> extends BaseType {
+
+    const dict<string, classname<NamedInputType>> BUILTINS = dict[
+        IntInputType::NAME => IntInputType::class,
+        StringInputType::NAME => StringInputType::class,
+        BooleanInputType::NAME => BooleanInputType::class,
```
ianhoffman, line 18: Would it make sense to add these to the schema's `INPUT_TYPES`? They should be surfaced during introspection, right?

ianhoffman (approved):

src/Types/Input/InputType.hack
jjergus: Good idea, I'll move it.

PR #66, opened 2021-05-14
Introspection

mwildehahn:
This is working for some cases. There is still more I need to implement (like fields/values etc.) but wanted to open up a PR for comments if this looked like the right direction.

This felt like a good balance between leveraging the existing codegen by defining the wrapper types in Introspection.hack while giving us more control over how objects expose their values.

This does introduce a lot of new classes/interfaces though I'm not sure if there is a way we can simplify that.

Getting the SDL via introspection will be easier if we have fragments: https://github.com/graphql-python/graphql-core/blob/03c2b06e4636ed8a89c7a4c7e56d244fd9d00bde/src/graphql/utilities/get_introspection_query.py#L24 but we can write the fragments out until we support them.


src/Introspection.hack
```diff
@@ -0,0 +1,289 @@
...
+    OBJECT = 'OBJECT';
+    INTERFACE = 'INTERFACE';
+    UNION = 'UNION';
+    ENUM = 'ENUM';
+    INPUT_OBJECT = 'INPUT_OBJECT';
+    LIST = 'LIST';
+    NON_NULL = 'NON_NULL';
+}
+
+<<\Slack\GraphQL\ObjectType('__Schema', 'Schema introspection')>>
```
mwildehahn, line 95: I used these wrapping types to prevent these fields from being queryable in the objects themselves. I'm not sure if there is a simpler way to do this. I'm going to look with some fresh eyes tomorrow 😄

src/Types/Input/NonNullableInputType.hack
```diff
@@ -0,0 +1,39 @@
...
+    final public function coerceNonVariableNode(Value\Value $node, dict<string, mixed> $variable_values): TInner {
+        return $this->inner_type->coerceNode($node, $variable_values);
+    }
+
+    <<__Override>>
+    public function assertValidVariableValue(mixed $value): TInner {
+        return $this->inner_type->assertValidVariableValue($value);
+    }
+
+    public function getTypeKind(): GraphQL\Introspection\__TypeKind {
```
mwildehahn, line 32: I needed to add these NonNullable types so we could output the NOT_NULL type kind.

src/Codegen/Generator.hack
```diff
@@ -200,9 +210,16 @@ class Field {
     protected function getArgumentInvocationStrings(): vec<string> {
         $invocations = vec[];
         foreach ($this->reflection_method->getParameters() as $index => $param) {
+            $type_text = $param->getTypeText();
+            // TODO: can we pass down dynamic context this way too?
+            if ($type_text === 'HH\classname<Slack\GraphQL\Introspection\IntrospectableSchema>') {
```
mwildehahn, line 215: We need to pass the schema down to inspect it. I added some special handling here if a function takes a schema object we'll pass it the schema. I thought this ended up being pretty nice and could let itself to supporting custom context objects. I'm open to other ways we can do this though.

src/Introspection.hack
```diff
@@ -0,0 +1,289 @@
+namespace Slack\GraphQL\Introspection;
+
+use namespace HH\Lib\Vec;
+
+interface IntrospectableSchema {
```
ianhoffman, line 5: If we're referencing the types in this file through the introspection namespace anyway, would it make sense to drop the `Introspectable` prefix from these names? Then they'd be referenced as `Introspection\Schema`, `Introspection\Type`, etc, which seems nice.

src/Types/Output/ObjectType.hack
```diff
@@ -42,4 +47,16 @@ abstract class ObjectType extends NamedOutputType {
 
         return $is_valid ? new GraphQL\ValidFieldResult($ret, $errors) : new GraphQL\InvalidFieldResult($errors);
     }
+
+    public function getFields(bool $include_deprecated = false): ?vec<GraphQL\Introspection\IntrospectableField> {
+        // TODO
```
ianhoffman, line 52: My feeling is that we should use `getFieldDefinition` to resolve fields for both introspection and normal querying. Is the plan to make `FieldDefinition` implement `IntrospectableField`?

src/Introspection.hack
```diff
@@ -0,0 +1,289 @@
...
+    public function getDescription(): ?string {
+        return $this->inner_type->getDescription();
+    }
+
+    <<\Slack\GraphQL\Field('type', 'Type of the field')>>
+    public function getType(): __Type {
+        return new __Type($this->inner_type->getType());
+    }
+
+    // TODO: Not sure how this is possible for input objects? Maybe if they're a class?
```
ianhoffman, line 284: Could maybe be done via an attribute.

src/Introspection.hack
```diff
@@ -0,0 +1,289 @@
+namespace Slack\GraphQL\Introspection;
```
ianhoffman, line 1: This does strike me as verbose, but I'm not sure that there's a simpler way to paper-over the difference between Input and Output types. For example, we want to be able to use both input and output enum types, scalar types, etc, interchangeably for the purpose of introspection (input and output enums are not two separate types). This lets us paper over that distinction by forcing both input and output enum types to implement `IntrospectableEnum`, presumably returning the same name and description for enums generated from the same Hack type.

Is there some other reason which I'm missing to use these interfaces?

src/Introspection.hack
```diff
@@ -0,0 +1,289 @@
...
+        if ($this->inner_type is IntrospectableObject || $this->inner_type is IntrospectableInterface) {
+            $fields = $this->inner_type->getFields($include_deprecated);
+            return $fields is nonnull ? Vec\map($fields, $field ==> new __Field($field)) : null;
+        }
+
+        return null;
+    }
+
+    <<\Slack\GraphQL\Field('interfaces', 'Interfaces the object implements, only applies to OBJECT')>>
+    public function getInterfaces(): ?vec<__Type> {
```
ianhoffman, line 152: What's behind the decision for these methods to have nullable returns types? (Instead of just returning an empty vec)

src/Types/Input/EnumInputType.hack
```diff
@@ -32,13 +28,13 @@ abstract class EnumInputType extends NamedInputType {
...
-                static::NAME,
-                \get_class($node)
-            );
+            throw new GraphQL\UserFacingError('Expected a valid value for %s, got %s', static::NAME, \get_class($node));
         }
         return $enum::getValues()[$value];
     }
 
+    public function getTypeKind(): GraphQL\Introspection\__TypeKind {
+        return GraphQL\Introspection\__TypeKind::ENUM;
```
ianhoffman, line 37: Should this just be a class constant exposes via a method?

src/Types/Input/NonNullableInputType.hack
```diff
@@ -0,0 +1,39 @@
...
+
+    <<__Override>>
+    public function coerceValue(mixed $value): TInner {
+        return $this->inner_type->coerceValue($value);
+    }
+
+    <<__Override>>
+    final public function coerceNonVariableNode(Value\Value $node, dict<string, mixed> $variable_values): TInner {
+        return $this->inner_type->coerceNode($node, $variable_values);
+    }
```
ianhoffman, line 25: This is sort of a bummer ... maybe a chance to introduce a trait (at some point) which automatically forwards calls along to the inner type?

src/Types/Input/Scalar/ScalarInputType.hack
```diff
@@ -0,0 +1,10 @@
+namespace Slack\GraphQL\Types;
+use namespace Slack\GraphQL;
+
+abstract class ScalarInputType extends NamedInputType {
```
ianhoffman, line 4: Should we call this `LeafInputType` to mirror `Output/LeafType.hack`?

src/Types/Output/InterfaceType.hack
```diff
@@ -0,0 +1,15 @@
...
+
+abstract class InterfaceType extends ObjectType implements GraphQL\Introspection\IntrospectableInterface {
+    <<__Override>>
+    final public function getTypeKind(): GraphQL\Introspection\__TypeKind {
+        return GraphQL\Introspection\__TypeKind::INTERFACE;
+    }
+
+    public function getPossibleTypes(): ?vec<GraphQL\Introspection\IntrospectableType> {
+        // TODO: need to return all the types that implement this interface.
+        // Should have a static property that we can codegen.
```
ianhoffman, line 12: +1

src/Types/Output/NonNullableOutputType.hack
```diff
@@ -0,0 +1,41 @@
...
+final class NonNullableOutputType<TInner, TResolved>
+    extends OutputType<TInner, TResolved>
+    implements GraphQL\Introspection\IntrospectableGeneric {
+
+    public function __construct(private OutputType<TInner, TResolved> $inner_type) {}
+
+    <<__Override>>
+    public function getName(): ?string {
+        return null;
+    }
```
ianhoffman, line 13: I don't think you need this?

src/Introspection.hack
mwildehahn: Yea, good call

src/Types/Output/ObjectType.hack
mwildehahn: Yup, this ended up working well

src/Introspection.hack
mwildehahn: We could have potentially made __Type, __Field etc interfaces and then had the generated types implement those. I think the issue with that though is then those fields will conflict with any fields defined in those objects. This way, none of the introspection fields are available on the generated types (aside from within the Query itself).

src/Introspection.hack
mwildehahn: This is from the spec: https://spec.graphql.org/June2018/#sec-Schema-Introspection. My guess is it is because only certain types implement this, so for those it should return an empty vec (ie object types with no interfaces) but for any other type it should be null since the property doesn't apply to it.

src/Types/Input/EnumInputType.hack
mwildehahn: Yea... I thought I did that at first but there was some reason I wanted a method. I can't remember.

mwildehahn:
Once https://github.com/mwildehahn/hack-graphql/pull/67 is merged I'll put that in and clean this up a bit. This doesn't implement all the features of introspection but it's enough and this is already really large 😄 . 

The remaining ones are returning interfaces a class implements, classes that implement an interface etc.

src/Types/Input/Scalar/ScalarInputType.hack
jjergus: leaf types are scalars + enums

src/Introspection.hack
jjergus: > we want to be able to use both input and output enum types, scalar types, etc, interchangeably for the purpose of introspection (input and output enums are not two separate types)

We could also make an (arbitrary) choice that either the input or the output type will implement the introspection interface, for the types that have both. It shouldn't really matter which; and it might save us some duplicate code, e.g. we won't have to implement things like getKind() on both.

src/Introspection.hack
jjergus: Yep something like this works:

```Hack
<<
  GraphQL\InputObjectType('Foo')
  GraphQL\DefaultValues(dict[
    'some_field' => 42,
    'other_field' => '...'
  ])
```

but also we don't have to support default values for now.

src/Introspection.hack
jjergus: > to prevent these fields from being queryable in the objects themselves

Hmm I don't quite understand this, what do you mean "queryable in the objects themselves"?

Having all these wrappers that mostly just delegate to the inner class really does seem like a lot of boilerplate code that should be avoidable somehow...

src/Types/Input/EnumInputType.hack
jjergus: If we want to put `<<GraphQL\Field>>` on these (either here or in the parent interface) -- which I think is worth trying, so we can avoid all those wrapper classes -- then we'll need these to be methods, since `<<GraphQL\Field>>` doesn't work on constants.

src/Types/Input/NonNullableInputType.hack
jjergus: Hmm we only need these wrappers for introspection, we should be able to avoid using them elsewhere (e.g. during execution). That could improve performance (less wrappers) and also simplify the code (e.g. we won't need these `coerce...` methods there because that's not something that introspection needs).

This would mean that there are 2 representations of each type, 1 for runtime (what we have now) and 1 for introspection, so e.g. FieldDefinition will have `getType()` that returns the normal type and `getIntrospectionType()` which returns the type wrapped in a NonNullable wrapper. So we pay for the simplified code here by making a few other places more complicated, but overall I think it should still end up being simpler...?

src/Codegen/Generator.hack
jjergus: Yeah this is a nice way to support context objects if needed, but I'm not sure we need it for the schema...couldn't these fields just do `new Schema()` instead of having the schema passed in? (or `Schema::get()` if we make it a singleton) The schema doesn't have any state that we'd need to pass around, or does it?

ianhoffman: > couldn't these fields just do new Schema() instead of having the schema passed in?

I guess the fear is that `Schema` doesn't exist until after codegen, so we'll be writing classes which reference a type that the type-checker can't necessarily resolve? Seems odd to have non-generated code reference generated code. But maybe I'm missing something?

src/Introspection.hack
mwildehahn: I was initially thinking we could do something like:

```
<<\Slack\GraphQL\InterfaceType('__Type', 'Introspectable type')>>
interface __Type {
...
}
```

with:

```
<<\Slack\GraphQL\ObjectType('User', 'User')>>
final class User implements __Type {
...
}
```

but then you'd be able to query:

```
{
    user {
       // introspection fields
       name
       kind
       fields
       ...etc
    }
}
```

which wouldn't make sense?

jjergus: OK so I think this is a bit confusing because the classes have the same name (only different namespace) but I think we're talking about 2 different classes:

- `\User` is the class that has `<<GraphQL\Object('User')>>` on it and has all the non-introspection field methods (`<<GraphQL\Field('favorite_color')>>` etc.)
- `\Slack\GraphQL\Test\Generated\User` is the class we generate from `\User`; this is the class that has all the introspection methods like `<<GraphQL\Field('kind')>>` (probably inherited from IntrospectableType, not actually part of the generated code itself)

mwildehahn:
I'm going to close out this PR. I have a v2 which is much simpler. The part I was missing was this: https://github.com/mwildehahn/hack-graphql/pull/66#discussion_r632842802. Generating interfaces for the introspection and having the generated types implement them worked much better!

Issue #27, opened 2021-05-10
Execution: fragments, @skip/@include


jjergus:
These are the tools that GraphQL provides for conditionally including specific fields in the response.

https://spec.graphql.org/draft/#sec-Field-Collection

Issue #26, opened 2021-05-10
Execution: interfaces, lists, non-null


jjergus:
Our executor/resolver is currently very basic, only supports simple object types and scalars. This issue is to implement most of the missing cases (they're all kind of intertwined so it's hard to do each one separately).

jjergus:
#29 implements lists and nullability, waiting for #23 before I start working on interfaces

PR #67, opened 2021-05-14
Refactor codegen

ianhoffman:
Refactors codegen pretty much entirely. We now have a set of builders for the various types we generate during codegen. These are:

* TypeBuilder: Base type for all builders which build types. Ensures the type has a name, a Hack type, a GQL type, and a superclass.
  * InputTypeBuilder: Base type for all builders which build input types. Ensures the type has a `TCoerced` type constant.
    * InputObjectBuilder: Builds a GQL Input Object from a Hack shape.
    * InputEnumBuilder: Builds an enum which can be used as input to a GQL field.
  * OutputTypeBuilder: Base type for all builders which build output types. Ensures the type has a `THackType` type constant.
    * ObjectBuilder: Builds GQL interface and objects from Hack interfaces, objects and shapes.
    * InputEnumBuilder: Builds an enum which can be used as output from a GQL field.
* FieldBuilder: Base type for builders which build GQL fields.
  * MethodFieldBuilder: Builds a GQL field from a Hack method.
  * StaticFieldBuilder: Builds a GQL field from a Hack static method.
  * ShapeFieldBuilder: Builds a GQL field from a Hack shape field.

All in all I'm pretty happy with this refactor given it removes code while breaking up a large file into many smaller files that each have their own boilerplate `use...` statements.


tests/gen/Generated.hack
```diff
@@ -4,193 +4,274 @@
  * To re-generate this file run vendor/bin/hacktest
```
ianhoffman, line 4: The diff here is just because I started sorting GQL classes at the end of `collectObjects()`.

jjergus: Would it make sense to put each generated class into a separate file? Then we won't need to sort them at all (we'll still need to sort some things like the dicts with input/output types though) + it might have other advantages, like fewer merge conflicts.

src/Codegen/Generator.hack
```diff
@@ -495,22 +241,32 @@ final class Generator {
...
+                    'Slack\\GraphQL\\Root',
+                    $mutation_fields
+                );
             }
 
             $objects = Vec\concat($top_level_objects, $objects, $inputs);
         }
 
-        return $objects;
+        return Vec\sort_by($objects, $object ==> $object->getGraphQLType());
```
mwildehahn, line 270: I think it's useful to keep `Query` and `Mutation` at the top of the file. I don't expect people to read the codegen very much but its a helpful guide to how the rest of it ties together.

mwildehahn (approved):
This looks great, this will help a lot with the introspection stuff. I've had to copy and paste the same method a bunch of places because we didn't have these base classes abstracted well enough. A couple small comments but let's merge this so I can pull it in.

src/Codegen/Builders/ObjectBuilder.hack
```diff
@@ -0,0 +1,69 @@
...
+use namespace HH\Lib\Vec;
+use type Facebook\HackCodegen\{CodegenClass, CodegenMethod, HackBuilderValues, HackCodegenFactory};
+
+
+/**
+ * Builder for GraphQL objects and interfaces.
+ *
+ * The annotated Hack type should be either a class, interface, or shape.
+ */
+// TODO: It probably makes sense to have a separate builder for interfaces.
```
jjergus, line 12: Yep I think once we implement proper execution for interfaces (I'm working on it as part of #26 and #27) that will also affect what we need to codegen for interfaces -- the overlap between object types and interfaces should actually be pretty small, e.g. interface fields shouldn't really have any resolvers, they should delegate to object types at runtime.

jjergus (approved):
This looks great, thanks!

I'm not sure if this applies anywhere right now, but if you notice any place where you needed to add multiple special cases for different types, sometimes it might be easier to fix the base classes instead of adding more special cases to the codegen -- e.g. I see we have all these type constants with different names (THackType, TCoerced, TExpected, TResolved) and some of these are really the same thing, so it might be easier in some cases to rename the constant, instead of adding a special case to the codegen to generate constants with different names, etc.

jjergus:
Oops another comment race condition...I guess it's a good sign that we both said the same things though :)

src/Codegen/Builders/ObjectBuilder.hack
ianhoffman: Oh interesting, I didn’t realize that. I was imagining it’d be the reverse: interface resolvers would contain common logic and objects would delegate to them if an interface field is being resolved. How does the interface resolver know which object resolver to delegate to? Would it need to introspect the Hack object?

src/Codegen/Generator.hack
ianhoffman: Will do. I like @jjergus ’s suggestion of breaking generated classes into separate files; will do that in a follow-up.

ianhoffman:
> This looks great, thanks!
> 
> I'm not sure if this applies anywhere right now, but if you notice any place where you needed to add multiple special cases for different types, sometimes it might be easier to fix the base classes instead of adding more special cases to the codegen -- e.g. I see we have all these type constants with different names (THackType, TCoerced, TExpected, TResolved) and some of these are really the same thing, so it might be easier in some cases to rename the constant, instead of adding a special case to the codegen to generate constants with different names, etc.

Yeah, good call out - I think an input type’s `TCoerced` and an output type’s `THackType` are really the same thing - the hack type from which the GQL type is generated. I’ll rename `TCoerced` to `THackType` in a subsequent PR! I still think it could be nice to have different base classes for input and output types just to make the type hierarchy explicit.

tests/gen/Generated.hack
ianhoffman: Will do in a follow up!

ianhoffman:
@mwildehahn I’m gonna merge this down so you can rebase against it. I’ll address feedback in a follow-up.

src/Codegen/Builders/ObjectBuilder.hack
jjergus: Yeah normally during execution, if we encounter a field whose return type is abstract (interface/union), then the first step when resolving the field is to narrow down its type to the correct subtype: https://spec.graphql.org/draft/#sel-JANRLFCJDDCAAEDBEB74H

This is needed for various things (e.g. for `__typename` to return the correct type name if it's included in the query; for `fragment Foo on TypeName` to correctly determine whether the fragment should be included or not...) but one consequence is that it is never needed to call any resolvers on an abstract type, since by the type we call any field resolver, we would have already narrowed it down to the correct non-abstract subtype.

jjergus:
> I think an input type’s TCoerced and an output type’s THackType are really the same thing - the hack type from which the GQL type is generated. I’ll rename TCoerced to THackType in a subsequent PR!

They're not exactly the same thing, but THackType seems like a reasonable description of both so that makes sense.

- values of input types are "coerced" at some point, which is a conversion from mixed -> TCoerced (could be called THackType)
- values of output types are "resolved" at some point, which is a conversion from TExpected -> TResolved (TExpected is also called THackType in some places)
- for types that are both input and output types, TCoerced = TExpected so I guess it makes sense to call these the same thing (e.g. THackType like you suggested)

ianhoffman:
> They're not exactly the same thing, but THackType seems like a reasonable description of both so that makes sense.
> 
> * values of input types are "coerced" at some point, which is a conversion from mixed -> TCoerced (could be called THackType)
> * values of output types are "resolved" at some point, which is a conversion from TExpected -> TResolved (TExpected is also called THackType in some places)
> * for types that are both input and output types, TCoerced = TExpected so I guess it makes sense to call these the same thing (e.g. THackType like you suggested)

Good point — I was playing a bit fast and loose with terminology. The third bullet point (combined with the fact that THackType = TExpected for output types) is what I was alluding to and why I think this rename makes sense.

src/Codegen/Builders/ObjectBuilder.hack
ianhoffman: Makes sense. It doesn't sound too difficult to implement a naive version of that spec.

I'm thinking about the case where the object type is a subclass of a Hack type which implements the interface. E.g., say I have the following object hierarchy:

```
<<GraphQL\Interface('Animal')>>
interface Animal {}

<<GraphQL\Object('Feline')>>
class Feline implements Animal {}

class Puma extends Animal {}
```

So in this case, say the field resolver for `Animal` receives an instance of `Puma`. We should be able to determine that `Puma` is an instance of `Feline` and so call the generated resolver for `Feline`, per the spec (there is no resolver for `Puma`). So what this means is: we need an algorithm which walks the object hierarchy of Puma, looking for the "closest" parent class which has a field resolver (i.e., is annotated with a GQL attribute). This sounds a lot like "[Method Resolution Order](https://www.geeksforgeeks.org/method-resolution-order-in-python-inheritance/)" in Python.

Anyway, we can probably do something more naive for a v1. Mostly thinking aloud here.

jjergus: We could make things a lot easier if we enforced some rules:

1. if `<<GraphQL\Object>>` is used on some class, there must not be `<<GraphQL\Object>>` on any of its subtypes or supertypes
2. ban `<<GraphQL\Object>>` on Hack interfaces (only classes)

1 is not strictly necessary but would make things easier

2 might actually be necessary, because if we have something like

```Hack
<<GraphQL\Object('A')>>
interface A { ... }

<<GraphQL\Object('B')>>
interface B { ... }

class C implements A, B { ... }
```

then if we have an instance of `C`, there is no way to decide if its GraphQL type should be A or B, unless we come up with some super-arbitrary rule like "first one in alphabetical order wins"

ianhoffman: Wouldn't (1) make it illegal to have both an interface and a concrete implementation of that interface in the same schema? Since if B implements A, then B cannot be annotated with `<<GraphQL\Object>>`. But this means you can never write a query for B instead of A. For example, I think this schema would be illegal: https://github.com/apollographql/apollo-tooling/blob/master/__fixtures__/starwars/schema.graphql

I agree that (2) is probably necessary. How do y'all do this at FB?

ianhoffman: Oh nvm, I see you're talking about `ObjectType`, not both `ObjectType` and `InterfaceType`, in (1). I guess that would help - we'd just need to find the single class of a class's superclasses which is annotated with `GraphQL\ObjectType`, and use that as the resolver.

For (2), another option would be to ensure the class only extends a single interface tagged with `GraphQL\InterfaceType` during codegen.

jjergus: > Wouldn't (1) make it illegal to have both an interface and a concrete implementation of that interface in the same schema?

The parent type would have to be a `<<GraphQL\Interface>>`, not `<<GraphQL\Object>>`.

> How do y'all do this at FB?

I think we ban all the ambiguous cases (so probably the 2 rules I mentioned apply).

> For (2), another option would be to ensure the class only extends a single interface tagged with GraphQL\InterfaceType during codegen.

In my example the class `C` is not annotated with any GraphQL attribute. Would the GraphQL codegen enforce the rule for every class in the codebase, whether or not it has a GraphQL attribute on it?

ianhoffman: We'd have to enforce it for any class implementing an interface tagged with a GQL attribute, yeah. Which sounds untenable for a large codebase.

PR #68, opened 2021-05-14
Respond to feedback on codegen PR

ianhoffman:
Responds to feedback on #67 

* Generate one file per class
* Rename `TCoerced` to `THackType` across all `InputType` implementations
* Use one file per class in the Codgen/Builders directory


src/Codegen/Generator.hack
```diff
@@ -25,63 +25,81 @@ final class Generator {
...
+        foreach ($handles as $handle) {
+            if ($handle->isDir()) {
+                \rmdir($handle->getRealPath());
+            } else {
+                \unlink($handle->getRealPath());
+            }
         }
+
+        \rmdir($directory);
+    }
```
ianhoffman, line 86: Deleting directories in PHP is such a pain. @jjergus , y'all should make a Hack standard library function for this. (Or is there one? I looked briefly but didn't see anything.)

src/playground/ErrorTestObj.hack
```diff
@@ -1,6 +1,6 @@
 use namespace Slack\GraphQL;
 
-<<GraphQL\ObjectType('ErrorTest', 'Test object for error handling')>>
+<<GraphQL\ObjectType('ErrorTestObj', 'Test object for error handling')>>
```
ianhoffman, line 3: This was very confusing for HackTest, so I renamed.

jjergus (approved):

PR #69, opened 2021-05-15
support optional field arguments

jjergus:
- this is very similar to optional input object fields (you can see I actually reused the `hasValue()` method)
- I got rid of `Graphpinator\Parser\Value\ArgumentValue` since it didn't add anything useful on top of `dict<string, Value\Value>`
- I removed the covariance from InputType::TCoerced (again) since covariant types can't be used for method arguments; because of that I had to add an `IInputType` interface that can be used in places where we don't know/don't care what TCoerced is (since `InputType<mixed>` is no longer a valid option)
- while I was at it I added a test case for ["list promotion"](https://spec.graphql.org/draft/#sec-List.Input-Coercion) too since I realized we weren't testing it anywhere. Our implementation doesn't actually match the examples in the cited spec, but those examples don't seem to be consistent with the text of the spec so I'm not sure what to do with it. Our current implementation is what seemed like the most logical option to me (it has the smallest amount of edge cases compared to other options). I should probably check what the reference implementation (graphql-js) does.


tests/gen/Generated.hack
```diff
@@ -481,31 +481,16 @@ final class Query extends \Slack\GraphQL\Types\ObjectType {
           ErrorTest::nonNullable(),
           async ($parent, $args, $vars) ==> \ErrorTestObj::getNonNullable(),
         );
-      case 'getConcrete':
```
ianhoffman, line 484: Hmm. Looks like we should make sort order here stable as well.

ianhoffman (approved):

tests/ArgumentTest.hack
```diff
@@ -0,0 +1,135 @@
...
+                dict['a' => 1, 'b' => 2, 'c' => 3],
+                dict[
+                    'a' => vec[69, 1, 1],
+                    'b' => vec[69, 2, 2],
+                    'c' => vec[69, 3, 3],
+                ],
+            ),
+
+            // Non-list values can get promoted to a list if used where a list is expected.
+            'list promotion' => tuple(
```
mwildehahn, line 107: TIL graphql supported this. This is cool!

mwildehahn (approved):

tests/gen/Generated.hack
jjergus: Good catch; I added a new commit here that does that. I only sorted fields generated from methods (object/interface/root types). The other cases (enums, input objects, shape-backed object types) have a clear natural order so it's probably better to preserve it.

PR #71, opened 2021-05-15
Introspection v2

mwildehahn:
Thanks for all the feedback on the previous PR. This is a second pass at introspection that is much simpler. Instead of having interfaces + wrapping types, we just have introspection interfaces with `<<GraphQL\InterfaceType(...)>>` attributes. The generated classes then implement these interfaces so when we're resolving the introspection query we read the fields from the classes themselves.

The codegen changes were pretty minimal since we can do most of the implementation of the interfaces within the base classes which is really nice. I haven't implemented all the introspection features yet, just enough to validate this direction.

If this looks good, I'll implement the remaining introspection features in another PR.


src/BaseSchema.hack
```diff
@@ -21,4 +25,24 @@ abstract class BaseSchema {
...
+    }
+
+    <<__Override>>
+    final public function getMutationType(): ?Introspection\__Type {
+        $mutation_type = static::MUTATION_TYPE;
+        return $mutation_type is nonnull ? $mutation_type::nonNullable() : null;
+    }
+
+    final public function getType(string $name): ?Introspection\__Type {
+        $types = Dict\merge(static::INPUT_TYPES, static::OUTPUT_TYPES);
```
mwildehahn, line 42: Should pull this out and memoize it

src/Codegen/Builders/Fields/MethodFieldBuilder.hack
```diff
@@ -44,7 +48,7 @@ class MethodFieldBuilder implements IFieldBuilder {
     }
 
     protected function getMethodCallPrefix(): string {
-        return '$parent->';
+        return Str\format('$parent%s', $this->reflection_method->isStatic() ? '::' : '->');
```
mwildehahn, line 51: I didn't end up needing this but this lets us use static methods if we want

src/FieldDefinition.hack
```diff
@@ -34,4 +35,16 @@ final class FieldDefinition<TParent, TRet, TResolved> implements IFieldDefinitio
...
         return await $this->type->resolveAsync($value, $field, $vars);
     }
+
+    public function getName(): string {
+        return $this->name;
+    }
+
+    public function getType(): Introspection\__Type {
+        if ($this->type is Types\NullableInputType<_> || $this->type is Types\NullableOutputType<_, _>) {
+            return $this->type;
```
mwildehahn, line 45: I wonder if instead of returning `$this->type` here we can just check:

```
if ($this->type is Introspection\IHasInnerType) {
   return $this->type->getInnerType();
}
```

and then drop `IntrospectNullableType`...

src/FieldDefinition.hack
```diff
@@ -34,4 +35,16 @@ final class FieldDefinition<TParent, TRet, TResolved> implements IFieldDefinitio
...
+        }
+
+        return new Introspection\NonNullable($this->type);
```
mwildehahn, line 48: We only need to output info for non nulls during introspection. With this we don't impact normal execution by wrapping classes unnecessarily with `NonNullable`, we just do it during introspection.

src/Resolver.hack
```diff
@@ -26,10 +26,15 @@ final class Resolver {
      * @see https://spec.graphql.org/draft/#sec-Execution
      */
     public async function resolve(
-        \Graphpinator\Parser\ParsedRequest $request,
+        string $input,
```
mwildehahn, line 29: Not really related, but we discussed not having clients need to understanding parsing the request so I implemented that here.

src/Codegen/Builders/TypeBuilder.hack
```diff
@@ -32,12 +32,18 @@ abstract class TypeBuilder<T as \Slack\GraphQL\__Private\GraphQLTypeInfo> {
...
     public function build(HackCodegenFactory $cg): CodegenClass {
-        return $cg
+        $class = $cg
             ->codegenClass($this->getGeneratedClassName())
             ->setIsFinal()
             ->setExtendsf('\%s', static::SUPERCLASS)
             ->addConstant($this->generateNameConstant($cg))
             ->addTypeConstant($this->generateTypeConstant($cg));
+
+        if ($this is ITypeWithFieldsBuilder) {
```
ianhoffman, line 42: It might be nicer to use a trait for this - to avoid putting this in the base class. Seems a bit odd for the base class to need to know about this.

src/FieldDefinition.hack
ianhoffman: That would result in unwrapping lists, right? Which I don't think is intended, since it should be possible to return a "kind" which is a list.

src/Introspection/IntrospectNullableType.hack
```diff
@@ -0,0 +1,22 @@
+namespace Slack\GraphQL\Introspection;
+
+/**
+* Types are nullable by default in GraphQL, however in this framework, we wrap
+* nullable types with either NullableInputType or NullableOutputType. This trait
+* lets us read introspection fields from the inner type since we don't need to
+* specify anything for nullable fields during introspection.
+*/
+trait IntrospectNullableType implements IHasInnerType {
```
ianhoffman, line 9: TIL traits can implement interfaces.

ianhoffman: I do wonder why we need both this and `IHasInnerType` though.

src/Introspection/QueryRootFields.hack
```diff
@@ -0,0 +1,16 @@
+namespace Slack\GraphQL\Introspection;
+
+use namespace Slack\GraphQL;
+
+final abstract class QueryRootFields {
+    <<GraphQL\QueryRootField('__schema', 'Schema introspection')>>
+    final public static function getSchema(): ?__Schema {
+        // TODO: switch to taking $schema as first param?
+        return new \Slack\GraphQL\Test\Generated\Schema();
```
ianhoffman, line 9: Yeah this seems... problematic. I don't know how we can get around this without injecting the schema as an argument.

src/Introspection/QueryRootFields.hack
```diff
@@ -0,0 +1,16 @@
...
+    }
+
+    <<GraphQL\QueryRootField('__type', 'Type introspection')>>
+    final public static function getType(string $name): ?__Type {
+        return (new \Slack\GraphQL\Test\Generated\Schema())->getType($name);
```
ianhoffman, line 14: could just do `return $this->getSchema()->getType()`.

src/Types/Input/ListInputType.hack
```diff
@@ -17,24 +18,35 @@ final class ListInputType<+TInner> extends InputType<vec<TInner>> {
...
+
+    <<__Override>>
+    final public function getKind(): GraphQL\Introspection\__TypeKind {
+        return GraphQL\Introspection\__TypeKind::LIST;
+    }
+
+    <<__Override>>
+    final public function getOfType(): GraphQL\Introspection\__Type {
+        if ($this->inner_type is NullableInputType<_>) {
+            return $this->inner_type;
```
ianhoffman, line 47: Should this be `return $this->inner_type->getInnerType()`?

src/Codegen/Builders/ObjectBuilder.hack
```diff
@@ -66,4 +68,9 @@ final class ObjectBuilder<TField as IFieldBuilder> extends OutputTypeBuilder<\Sl
 
         return $method;
     }
+
+    final public function getFieldNames(): keyset<string> {
+        return Vec\map($this->fields, $field ==> $field->getName())
```
ianhoffman, line 73: `Keyset\map`?

ianhoffman:
This looks really good and reasonable to me. Curious what @jjergus thinks!

ianhoffman (approved):

src/FieldDefinition.hack
mwildehahn: Yea, `IHasInnerType` would need to be renamed to make it only apply to nullables since those are the ones we just want to skip to the inner type. This would let us skip to the inner types for introspection but we still need to define `getName` and `getKind` on `NullableOutputType` and `NullableInputType` since they extend from `BaseType`. It seems weird to have to define those even though they'll never be called 🤔 .

The current approach seems easier to follow.

src/Introspection/IntrospectNullableType.hack
mwildehahn: I needed a way to access `inner_type` on NullableInput and NullableOutput, however both those are different types: `InputType<TInner>` and `OutputType<TInner, TResolved>`.

I tried setting:

```
trait IntrospectNullableType implements IHasInnerType {
    protected __Type $inner_type;
    ...
}
```

But then when we use the trait in the nullable types it complains about a type mismatch. The interface lets both of those methods return inner type as __Type and for us to access that in this trait.

src/FieldDefinition.hack
ianhoffman: You could also write a `getTypeForIntrospection()` method on `BaseType` and then just call it here. Then `NullableInputType` and `NullableOutputType` could override it to return `$this`, while the base implementation could return `new Introspection\NonNullableType($this->type)`.

 This is fine though.

ianhoffman: Also wondering if it might make sense for `getName()` and `getKind()` on nullable types to automatically forward to the inner type? Then instead of needing to unwrap these types, maybe you could just call these methods directly.

And finally, why are we returning `$this->type` if the type is nullable? Don't we want to return `$this->type->getInnerType()` since the nullable wrapper type shouldn't be surfaced by the introspection system?

src/Introspection/IntrospectNullableType.hack
ianhoffman: Hmm. I think you can make it work with the following:

```hack
trait IntrospectNullableType<+T as __Type> {
    private T $inner_type;

    protected function getInnerType(): __Type {
        return $this->inner_type;
    }

    // etc
}

final class NullableInputType<+TInner as nonnull> extends InputType<?TInner> {
    use GraphQL\Introspection\IntrospectNullableType<InputType<TInner>>;

    // etc
}

final class NullableOutputType<TInner, TResolved> extends OutputType<?TInner, ?TResolved> {
    use GraphQL\Introspection\IntrospectNullableType<OutputType<TInner, TResolved>>;

    // etc
}
```

src/BaseSchema.hack
jjergus: Why not just

```Hack
$type = static::INPUT_TYPES[$name] ?? static::OUTPUT_TYPES[$name] ?? null;
```

instead of constructing the big map?

src/FieldDefinition.hack
jjergus: What if we had `getType()` and `getIntrospectionType()` as 2 different methods here? I haven't really thought it through very much but it feels like that might give us more type safety (e.g. we won't accidentally return a NonNullable-wrapped type during execution/validation; or accidentally return a Nullable-wrapped type for introspection)

src/Introspection/QueryRootFields.hack
jjergus: It might be better to avoid using GraphQL\QueryRootField + codegen for these 2 fields. Instead, we could add a bit of extra codegen code to the generator for the Query root type, which would add these 2 fields to the generated output (so the generated output would be all the normal root fields parsed from various files + these 2 hardcoded fields).

src/Introspection/__EnumValue.hack
```diff
@@ -0,0 +1,19 @@
+namespace Slack\GraphQL\Introspection;
+
+use namespace Slack\GraphQL;
+
+<<GraphQL\InterfaceType('__EnumValue', 'Enum value introspection')>>
+interface __EnumValue {
```
jjergus, line 6: Some of these could just be classes instead of interfaces, e.g. I don't think we'll need a non-introspection version of EnumValue that would implement this interface.

src/Introspection/__EnumValue.hack
```diff
@@ -0,0 +1,19 @@
+namespace Slack\GraphQL\Introspection;
+
+use namespace Slack\GraphQL;
+
+<<GraphQL\InterfaceType('__EnumValue', 'Enum value introspection')>>
```
jjergus, line 5: These should be object types, not interfaces in GraphQL. If we're not allowing `<<GraphQL\ObjectType>>` on Hack interfaces, we might have to allow an exception for these introspection types.

src/Types/BaseType.hack
```diff
@@ -1,13 +1,32 @@
...
 
-// TODO: Implement introspection: https://spec.graphql.org/draft/#sec-Introspection
-// It may or may not be a good idea to reuse the existing codegen to generate the introspection types.
-// If we did that, then this would be:
-// <<GraphQL\Object('__Type', '')>>
-abstract class BaseType {
-    // <<GraphQL\Field('name', '')>>
+use namespace Slack\GraphQL;
+
+abstract class BaseType implements GraphQL\Introspection\__Type {
```
jjergus, line 5: If you try my suggestion from above (have separate `getType()` and `getIntrospectionType()` methods on field definition etc.) you'd probably want to remove the `implements GraphQL\Introspection\__Type` and add it to all subtypes except `Nullable...` -- that would ensure that we don't return an invalid kind of wrapper at runtime/during introspection

Actually it might be even better if we put `implements GraphQL\Introspection\__Type` only on 1 copy of each type that has input+output version, that would prevent us from e.g. returning duplicate types from `$schema->getTypes()`

src/BaseSchema.hack
mwildehahn: Ah good point, will fix this

src/Introspection/QueryRootFields.hack
mwildehahn: Yea, I'll leave this like this for this PR and change in a follow up since this is pretty straight forward

ianhoffman: Oooh, that makes a lot of sense.

src/Introspection/__EnumValue.hack
mwildehahn: Yea that makes sense. I haven't implemented anything with enums yet and will do that in a follow up PR

src/Types/BaseType.hack
mwildehahn: I like this, let me try it out

src/Introspection/IntrospectNullableType.hack
mwildehahn: I didn't even think to use generics, great call.

jjergus: You probably don't need generics for introspection types, since we don't do thing like coercion/field resolving that we would want to write in a type-safe way.

src/Codegen/Builders/ObjectBuilder.hack
mwildehahn: TIL

mwildehahn:
Updated based on the reviews

src/FieldDefinition.hack
```diff
@@ -34,4 +35,16 @@ final class FieldDefinition<TParent, TRet, TResolved> implements IFieldDefinitio
...
+    public function getName(): string {
+        return $this->name;
+    }
+
+    public function getIntrospectionType(): Introspection\__Type {
+        if ($this->type is Introspection\INullableType) {
+            return $this->type->getInnerType();
+        }
+
+        return new Introspection\NonNullable($this->type as Introspection\__Type);
```
jjergus, line 48: Do we need some kind of recursion here if we're dealing with nested lists? e.g. `IntInputType::nonNullable()->nullableListOf()` needs removing the Nullable wrapper from the list and adding a NonNullable wrapper around the inner int type

src/Introspection/__EnumValue.hack
jjergus: Some of these could even be shapes maybe, now that we support generating object types from shapes.

```
<<GraphQL\Object('__EnumValue')>>
type __EnumValue = shape(
  'name' => string,
  // ...
);
```

jjergus (approved):
This looks good to me, but if the wrapping/unwrapping/rewrapping keeps getting more complicated, or you run into more similar complications as you're implementing the remaining features, I wonder if we should revisit whether it's still a good idea to reuse the runtime classes for introspection (I know it was probably me who originally suggested it so I'm arguing against myself here): How much did we get out of it and how much it costs us?

Right now it looks like we're reusing a few things like `getName()` which is used both at runtime and for introspection; few other things like `getKind()` are neutral (we don't save anything but it also doesn't cost us anything whether `getKind()` is on the runtime class or a separate class/shape) and a few things are downsides (all the type rewrapping). I expect we'll probably end up reusing a few more things when validation is implemented, but also more complications might show up, so it's probably worth revisiting over time.

mwildehahn:
Merging this so we can start iterating with validation etc.

PR #73, opened 2021-05-17
type-safe literals

jjergus:
Adds parser nodes for `IntLiteral`, `StringLiteral` etc. This makes some of the coercion code simpler and safer, and also improves some of our existing error messages (see updated tests).

I removed `getRawValue()` from `Parser\Value` (only `Literal` values have it now). The method would throw if there is a variable reference nested anywhere inside the literal, which is dangerous and would probably surprise anyone who calls it.

### DIFF
--- a/src/Graphpinator/Parser/Value/Literal.hack
+++ b/src/Graphpinator/Parser/Value/Literal.hack
@@ -1,14 +1,15 @@
 namespace Graphpinator\Parser\Value;
 
-final class Literal implements \Graphpinator\Parser\Value\Value {
+abstract class Literal implements \Graphpinator\Parser\Value\Value {
+    abstract const type THackType;
 
-    public function __construct(private mixed $value) {}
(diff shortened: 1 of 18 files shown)


src/Graphpinator/Parser/Value/Literal.hack
```diff
@@ -1,14 +1,15 @@
...
 
-    public function getRawValue(): mixed {
+    final public function getRawValue(): this::THackType {
         return $this->value;
     }
 
-    public function accept(ValueVisitor $valueVisitor): mixed {
+    final public function accept(ValueVisitor $valueVisitor): mixed {
         return $valueVisitor->visitLiteral($this);
     }
```
ianhoffman, line 14: Unrelated to this PR, but pretty sure we can remove these.

ianhoffman (approved):

mwildehahn (approved):

PR #76, opened 2021-05-18
Handle parse errors

ianhoffman:


src/Resolver.hack
```diff
@@ -26,25 +26,28 @@ final class Resolver {
...
-            try {
-                list($ret['data'], $errors) = await $this->resolveImpl($request, $variables, $operation_name);
-            } catch (UserFacingError $e) {
-                $errors = vec[$e];
-            } catch (\Throwable $e) {
-                // TODO: This shoud not happen; if it does, it's a bug in the GraphQL framework. Every exception should
-                // either be UserFacingError, or caught somewhere. However, we probably still want to catch arbitrary
-                // exceptions here just in case and return *some* reasonable response.
-                throw $e; // for now, so as to not break existing tests
+        if ($request is nonnull) {
```
jjergus, line 39: style nit: This would be a bit easier to read (and maybe safer) with early returns instead of these nested ifs

```Hack
try {
  // parse
} catch (...) {
  return ...
}

// no if needed here
try {
  // validate
} catch (...) {
  return ...
}

// no if needed
try {
  // resolve
} catch (...) {
  ...
}
```

src/Resolver.hack
```diff
@@ -55,7 +58,26 @@ final class Resolver {
         return $ret;
     }
 
-    public async function resolveImpl(
+    private function parseRequest(string $input): \Graphpinator\Parser\ParsedRequest {
+        $source = new \Graphpinator\Source\StringSource($input);
+        $parser = new \Graphpinator\Parser\Parser($source);
+        try {
+            return $parser->parse();
+        } catch (\Graphpinator\Exception\GraphpinatorBase $e) {
```
jjergus, line 66: Would it be hard to update the tokenizer+parser to throw UserFacingErrors so we don't need to transform them here?

jjergus (approved):

src/Resolver.hack
ianhoffman: Yeah, I thought about doing this. It's certainly non-trivial and the payoff didn't seem large, so I punted for now, but I do think we should do it eventually. How about I file an issue?

src/Resolver.hack
ianhoffman: Yeah, I like that. Will update.

src/Resolver.hack
```diff
@@ -26,25 +26,28 @@ final class Resolver {
...
-                list($ret['data'], $errors) = await $this->resolveImpl($request, $variables, $operation_name);
-            } catch (UserFacingError $e) {
-                $errors = vec[$e];
-            } catch (\Throwable $e) {
-                // TODO: This shoud not happen; if it does, it's a bug in the GraphQL framework. Every exception should
-                // either be UserFacingError, or caught somewhere. However, we probably still want to catch arbitrary
-                // exceptions here just in case and return *some* reasonable response.
-                throw $e; // for now, so as to not break existing tests
+        if ($request is nonnull) {
+            $validator = new \Slack\GraphQL\Validation\Validator($this->schema);
```
mwildehahn: So cool that we got this in already!

mwildehahn (approved):

src/Resolver.hack
ianhoffman: Filed: https://github.com/mwildehahn/hack-graphql/issues/77

PR #78, opened 2021-05-18
introspection shapes

jjergus:
(doesn't work yet)

Here I tried to prototype what it would look like if we used simple shapes for introspection. We probably still need a class for __Type, because

- some fields have an argument
- we need a separate class for "type declaration" and "type reference", otherwise we could end up with infinite nested shapes (because an object type may have a field that returns the same type, e.g. "best_friend" on User returns User)


ianhoffman:
I like the look of this - seems to reduce boilerplate somewhat. It'd be helpful to see where we're constructing these shapes, type definitions and references during introspection.

src/Introspection/Type/NamedTypeDeclaration.hack
```diff
@@ -0,0 +1,78 @@
+namespace Slack\GraphQL\Introspection;
+
+use namespace HH\Lib\Vec;
+
+final class NamedTypeDeclaration extends __Type {
```
mwildehahn, line 5: Would would we get these declarations back from generated classes?

mwildehahn:
I took a crack yesterday at separating out introspection types from runtime types: https://github.com/mwildehahn/hack-graphql/pull/81. The part I haven't figured out, and may have gone in the wrong direction on, is I was trying to generate separate objects from the runtime objects so we didn't have to overload field definition or deal with runtime wanting nulls to work one way and introspection wanting them to work another way. We'd generate something like this for a type:

```
namespace Slack\GraphQL\Test\Generated\Introspection;
use namespace Slack\GraphQL;
use namespace Slack\GraphQL\Introspection;

final class Concrete extends \Slack\GraphQL\Introspection\V2\ObjectType {

  const NAME = 'Concrete';
  const ?string DESCRIPTION = null;

  <<__Override>>
  public function getFields(): vec<\Slack\GraphQL\Introspection\V2\__Field> {
    return vec[
      \Slack\GraphQL\Introspection\V2\__Field::for(
        'bar',
        Introspection\V2\StringType::nonNullable(),
      )
      ,
      \Slack\GraphQL\Introspection\V2\__Field::for(
        'baz',
        Introspection\V2\StringType::nonNullable(),
      )
      ,
      \Slack\GraphQL\Introspection\V2\__Field::for(
        'foo',
        Introspection\V2\StringType::nonNullable(),
      )
      ,
    ];
  }
}
```

This gives us the same API as defining lists/nullables etc. but just returns the simpler introspection types. The thing I haven't looked at is actually resolving these in the query. We'd need to write a `resolveAsync` method on `\Slack\GraphQL\Introspection\V2\ObjectType`.

jjergus:
> How would we get these declarations back from generated classes?

I added a second commit here that demonstrates one way to do this (you probably want to look at the changes to the codegen output in the commit first). It adds an `introspect` method to all the generated types that returns the introspection shape representing that type. It takes a `__Schema` as an argument, which can be used to resolve `NamedTypeReference`s as needed.

jjergus:
https://github.com/mwildehahn/hack-graphql/pull/85

PR #79, opened 2021-05-18
interface base class + codegen + execution

### LINKED ISSUE #26: Execution: interfaces, lists, non-null

jjergus:
Closes #26

This adds a separate base class for interfaces + codegen to generate them. The generated `resolveAsync` figures out the concrete object type for the provided `$value`, then delegates to `resolveAsync` on the respective `ObjectType` class.

This means that `FieldDefinition`s on interface types no longer need a `$resolver`, but I haven't changed that yet so as to not change too many things in a single PR.

Right now the only effect of this is that we can correctly resolve the [`__typename` meta-field](https://spec.graphql.org/draft/#sec-Type-Name-Introspection) (see updated test) but this will be very important when executing queries with fragments.


src/Codegen/Builders/CompositeBuilder.hack
```diff
@@ -0,0 +1,88 @@
...
+    CodegenClassConstant,
+};
+
+
+/**
+ * Builder for GraphQL objects and interfaces (has a subclass for each).
+ *
+ * The annotated Hack type should be either a class, interface, or shape.
+ */
+abstract class CompositeBuilder<TField as IFieldBuilder>
```
jjergus, line 18: This is moved from ObjectBuilder; I did `git mv` but it seems that Git lost that information after I made changes to the file :(

The only changes here are: class name, docblock, removed SUPERTYPE const

src/Codegen/Builders/CompositeBuilder.hack
```diff
@@ -0,0 +1,88 @@
...
+
+    public function __construct(
+        \Slack\GraphQL\__Private\CompositeType $type_info,
+        string $hack_type,
+        protected vec<TField> $fields,
+    ) {
+        parent::__construct($type_info, $hack_type);
+    }
+
+    final public static function fromTypeAlias<T>(
```
ianhoffman, line 29: Should this live on `ObjectBuilder`? Or will we support building interfaces from shapes?

src/Types/Output/ObjectType.hack
```diff
@@ -30,6 +30,10 @@ abstract class ObjectType extends NamedOutputType implements GraphQL\Introspecti
         $results = await Dict\map_async(
             $child_fields,
             async $child_field ==> {
+                if ($child_field->getName() === '__typename') {
+                    // https://spec.graphql.org/draft/#sec-Type-Name-Introspection
+                    return new GraphQL\ValidFieldResult(static::NAME);
```
ianhoffman, line 35: nice

src/Codegen/Generator.hack
```diff
@@ -225,12 +225,32 @@ final class Generator {
...
+                $graphql_attribute = $rc->getAttributeClass(\Slack\GraphQL\InterfaceType::class);
+                if ($graphql_attribute is nonnull) {
+                    $objects[] = new InterfaceBuilder(
+                        $graphql_attribute,
+                        $rc->getName(),
+                        $fields,
+                        $hack_class_to_graphql_object,
+                    );
+                }
+                $graphql_attribute = $rc->getAttributeClass(\Slack\GraphQL\ObjectType::class);
```
ianhoffman, line 253: Can a class be tagged with both InterfaceType and ObjectType?

ianhoffman (approved):

src/Types/Output/InterfaceType.hack
```diff
@@ -0,0 +1,31 @@
...
+        GraphQL\Variables $vars,
+    ): Awaitable<GraphQL\FieldResult<dict<string, mixed>>>;
+
+    <<__Override>>
+    final public function getKind(): GraphQL\Introspection\__TypeKind {
+        return GraphQL\Introspection\__TypeKind::INTERFACE;
+    }
+
+    <<__Override>>
+    final public function getFields(): vec<GraphQL\Introspection\__Field> {
```
ianhoffman, line 28: If we make `CompositeType` an abstract class extending `NamedOutputType`, we could move `getFields()` and the abstract definition of `getFieldDefinition` into it for both `InterfaceType` and `ObjectType`. A very small win?

src/Validation/Rules/FieldsOnCorrectTypeRule.hack
```diff
@@ -5,6 +5,11 @@ final class FieldsOnCorrectTypeRule extends ValidationRule {
     <<__Override>>
     public function enter(\Graphpinator\Parser\Node $node): void {
         if ($node is \Graphpinator\Parser\Field\Field) {
+            if ($node->getName() === '__typename') {
```
ianhoffman, line 8: Thanks!

mwildehahn (approved):

src/Codegen/Builders/CompositeBuilder.hack
jjergus: Good catch, I don't think we can support shape-backed interfaces because we would have no way to map a value to the correct subtype at runtime (the exact thing I'm adding in this PR for classes). I'll move it.

src/Codegen/Generator.hack
jjergus: I guess it could but it doesn't make a lot of sense (we'd have an interface type with a single concrete subtype). Probably better to forbid it, at least for now.

src/Types/Output/InterfaceType.hack
jjergus: I'm planning to clean up the FieldDefinition stuff (including removing the resolvers from interface FieldDefinitions) but probably better to wait until the introspection/validation PRs are settled first.

PR #80, opened 2021-05-18
Codegen argument defs

ianhoffman:
Each field definition should know about its parameters. This will be useful for validation and introspection.

For now, we just generate shapes containing the type, name, and default value for each argument. This might get fancier if we need it to.

### DIFF
--- a/src/FieldDefinition.hack
+++ b/src/FieldDefinition.hack
@@ -1,5 +1,13 @@
 namespace Slack\GraphQL;
 
+type ArgumentDefinition = shape(
+    'name' => string,
+    'type' => Types\IInputType,
+    ?'description' => string,
+    ?'default_value' => mixed,
+    ?'deprecation_reason' => string,
+);
+
 interface IFieldDefinition extends Introspection\__Field {
     public function getName(): string;
     public function getType(): Types\IOutputType;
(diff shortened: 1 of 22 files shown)


mwildehahn (approved):
You're now a codegen expert 😉

ianhoffman:
> You're now a codegen expert 😉

Hah. I could have sworn that there's a nicer way to do the codegen in this PR — maybe using something like `HackBuilderValues::shapeWithPerKeyRendering`, but I couldn't quite figure it out and so just went with this.

src/FieldDefinition.hack
```diff
@@ -1,5 +1,13 @@
 namespace Slack\GraphQL;
 
+type ArgumentDefinition = shape(
+    'name' => string,
+    'type' => Types\IInputType,
+    ?'description' => string,
+    ?'default_value' => mixed,
+    ?'deprecation_reason' => string,
+);
```
jjergus, line 9: Depending on whether we end up reusing any of this for introspection, we may not need most of these here (name is redundant because we have the names as dict keys, only type and `'has_default' => bool` would be needed for validation)

ianhoffman: Good point, I'll remove most of this in a subsequent PR.

PR #82, opened 2021-05-19
Port KnownArgumentNamesRule

ianhoffman:
Ports the [KnownArgumentNamesRule](https://github.com/graphql/graphql-js/blob/main/src/validation/rules/KnownArgumentNamesRule.js) (not accounting for directives).

I had to add back the `Graphpinator\Parser\Value\ArgumentValue` AST node since this rule should only execute against arguments, not arbitrary value nodes. @jjergus , let me know if there's a better way to do this.


src/__Private/Utils/Stack.hack
```diff
@@ -26,6 +25,6 @@ final class Stack<T> {
     }
 
     public function asVec(): vec<T> {
-        return Dict\sort_by_key($this->items) |> vec($$);
+        return Vec\take($this->items, $this->length);
```
ianhoffman, line 28: Whoops!

src/Validation/Rules/KnownArgumentNamesRule.hack
```diff
@@ -0,0 +1,25 @@
+namespace Slack\GraphQL\Validation;
+
+
+final class KnownArgumentNamesRule extends ValidationRule {
+    <<__Override>>
+    public function enter(\Graphpinator\Parser\Node $node): void {
+        if ($node is \Graphpinator\Parser\Value\ArgumentValue) {
```
jjergus, line 7: > I had to add back the Graphpinator\Parser\Value\ArgumentValue AST node since this rule should only execute against arguments, not arbitrary value nodes. @jjergus , let me know if there's a better way to do this.

You could do something like

```Hack
if ($node is Parser\Field) {
  foreach ($node->getArguments() as $name => $value) {
    ...
```

here but your solution is good too. I only removed it so that I could reuse the same coercion method for arguments and input object fields -- and it looks like you addressed that by adding `getArgumentValues`.

jjergus (approved):

src/Validation/Rules/KnownArgumentNamesRule.hack
ianhoffman: Yeah, I think it's nice for the validation rules to not need to do too much AST traversal themselves. Since then they'd need to manage keeping `TypeInfo` in sync too.

PR #83, opened 2021-05-19
Port KnownTypeNamesRule

ianhoffman:
Port [KnownTypeNamesRule](https://github.com/graphql/graphql-js/blob/main/src/validation/rules/KnownTypeNamesRule.js).

We don't handle SDL validation because we'd like to that separately during codegen.

### DIFF
--- a/src/Validation/Rules/KnownTypeNamesRule.hack
+++ b/src/Validation/Rules/KnownTypeNamesRule.hack
@@ -1,0 +1,14 @@
+namespace Slack\GraphQL\Validation;
+
+final class KnownTypeNamesRule extends ValidationRule {
+    <<__Override>>
+    public function enter(\Graphpinator\Parser\Node $node): void {
+        if ($node is \Graphpinator\Parser\TypeRef\NamedTypeRef) {
+            $schema = $this->context->getSchema();
+            $type = (new $schema())->getIntrospectionType($node->getName());
+            if ($type is null) {
+                $this->reportError($node, 'Unknown type "%s".', $node->getName());
+            }
+        }
+    }
+}
--- a/src/BaseSchema.hack
+++ b/src/BaseSchema.hack
@@ -3,6 +3,7 @@ namespace Slack\GraphQL;
 use namespace HH\Lib\Dict;
 
 // TODO: this should be private
+<<__ConsistentConstruct>>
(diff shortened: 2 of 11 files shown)


src/Validation/Rules/KnownTypeNamesRule.hack
```diff
@@ -0,0 +1,14 @@
+namespace Slack\GraphQL\Validation;
+
+final class KnownTypeNamesRule extends ValidationRule {
+    <<__Override>>
+    public function enter(\Graphpinator\Parser\Node $node): void {
+        if ($node is \Graphpinator\Parser\TypeRef\NamedTypeRef) {
+            $schema = $this->context->getSchema();
+            $type = (new $schema())->getIntrospectionType($node->getName());
```
ianhoffman, line 8: What do y'all think about always making the schema be an object (i.e., non-static)? It would be nice to not have to juggle between static and non-static methods.

jjergus: Seems reasonable, maybe a singleton like all the type classes?

`SchemaClass::get()->whatever()`

jjergus (approved):
Do we want separate rules for all of these?

- type name is valid
- variable type is input type
- fragment type is composite type

Or just 2 rules?

- variable type is valid and it's an input type
- fragment type is valid and it's a composite type

src/Validation/Rules/KnownTypeNamesRule.hack
ianhoffman: How would `get()` know to return our generated schema? Unless `SchemaClass` is itself generated — but then how would this method know which `SchemaClass` to reference? We can't control the namespace under which the generated code will live.

Unless I'm missing something, I think it's easier to just pass the schema down from the resolver.

ianhoffman:
> Do we want separate rules for all of these?
> 
> * type name is valid
> * variable type is input type
> * fragment type is composite type
> 
> Or just 2 rules?
> 
> * variable type is valid and it's an input type
> * fragment type is valid and it's a composite type

Good question. The reference implementation has separate rules for these three cases. I think I'm inclined to keep it that way? These rules will likely grow in complexity so having each be as narrowly-scoped as possible seems good.

mwildehahn (approved):

PR #84, opened 2021-05-19
clean up introspection a bit

jjergus:
Alternative to #78

This reuses the existing classes for introspection, but tries to avoid the downsides:

- input types (except input objects) delegate to their respective output types, to avoid code duplication
- we have separate wrappers for introspection, Introspection\ListType and Introspection\NonNullableType which makes them less confusing then when we tried to reuse the existing incompatible ones
- we have `introspect()` declared on BaseType, implemented on the relevant subclasses, which gives us a unified API to get the introspection version of any type (we don't need separate re-wrapping logic on FieldDefinition and other places)


src/Types/Input/NullableInputType.hack
```diff
@@ -36,4 +34,13 @@ final class NullableInputType<TInner as nonnull>
     public function assertValidVariableValue(mixed $value): ?TInner {
         return $value is null ? null : $this->inner_type->assertValidVariableValue($value);
     }
+
+    <<__Override>>
+    final public function introspect(): GraphQL\Introspection\__Type {
+        $inner = $this->inner_type->introspect();
+        while ($inner is GraphQL\Introspection\NonNullableType) {
```
mwildehahn, line 41: This is a little hard to follow: if we're introspecting a nullable type, drill down till the first type that isn't nonnullable.

jjergus: There should never be more than but since we don't statically enforce it I thought this would be safer.

jjergus:
replaced by https://github.com/mwildehahn/hack-graphql/pull/85

PR #85, opened 2021-05-20
unify input/output types

jjergus:
(the 2 commits can be reviewed separately)

This merges all the types into a single unified hierarchy.

Input/output-ness of a type is represented by implementing one (or both!) of the interfaces `I[Named]InputType` or `I[Named]OutputType` (usually not directly but using the respective trait `T[Named][Input/Output]Type`, since we needed a place for the shared logic and interfaces can't have code).

```
BaseType
 |- NamedType
 |   |- InputObjectType     (implements INamedInputType)
 |   |- CompositeType       (implements INamedOutputType)
 |   |   |- InterfaceType
 |   |   `- ObjectType
 |   `- LeafType            (implements INamedInputType, INamedOutputType)
 |       |- EnumType
 |       `- ScalarType
 |           |- BooleanType
 |           |- IntType
 |           `- StringType
 |- ListInputType           (implements IInputType)
 |- NullableInputType       (implements IInputType)
 |- ListOutputType          (implements IOutputType)
 `- NullableOutputType      (implements IOutputType)
```

The main problem here is that if a type is both an input and an output type, does e.g. `->nullableListOf()` return an InputList or OutputList? I solved it by having 2 methods (`nullableListOfI()` and `nullableListOfO()` -- could use a better name I think), which is not great but I guess it's better to have 2 methods than 2 copies of each class...?

Also obviously there's multiple-inheritance now (via interfaces/traits) which always makes things more confusing.

There's more cleanups that could be done if we go this way (e.g. introspection stuff) but I wanted to give you a chance to review this first, before I spend too much time on it.


jjergus:
I added a 2nd commit where I updated the introspection implementation to no longer use any re-wrapping -- see the TNullable and TNonNullable traits for how we manage to return the correct data without the re-wrapping. It is a bit confusing, but at least the confusion is handled in that one place and no one needs to worry about it elsewhere -- now every BaseType instance can be used as an introspection type and will return the correct data.

jjergus:
The 2nd commit adds 2 more interfaces, which is not great but they are not pure overhead and can be used for useful stuff -- e.g. the nullable wrappers now guarantee to be wrapping a non-nullable type.

Basically each type now implements one or both of Input/OutputType, and exactly one of Nullable/NonNullableType.

```
BaseType
 |- NamedType               (implements INonNullableType)
 |   |- InputObjectType     (implements INamedInputType)
 |   |- CompositeType       (implements INamedOutputType)
 |   |   |- InterfaceType
 |   |   `- ObjectType
 |   `- LeafType            (implements INamedInputType, INamedOutputType)
 |       |- EnumType
 |       `- ScalarType
 |           |- BooleanType
 |           |- IntType
 |           `- StringType
 |- ListInputType           (implements INonNullableType, IInputType)
 |- NullableInputType       (implements INullableType,    IInputType)
 |- ListOutputType          (implements INonNullableType, IOutputType)
 `- NullableOutputType      (implements INullableType,    IOutputType)
```

ianhoffman:
> I added a 2nd commit (can be reviewed separately) where I updated the introspection implementation to no longer use any re-wrapping -- see the TNullable and TNonNullable traits for how we manage to return the correct data without the re-wrapping. It is a bit confusing, but at least the confusion is handled in that one place and no one needs to worry about it elsewhere -- now every BaseType instance can be used as an introspection type and will return the correct data.

Clever! This seems like a nice way to preserve type safety and still avoid the clunky re-wrapping in non-nullable introspection-only types we saw in previous implementations.

The actual code is a bit confusing, but as you say, at least it's all in one place.

src/Introspection/__Type.hack
```diff
@@ -10,31 +10,31 @@ use Slack\GraphQL;
 interface __Type {
 
     <<GraphQL\Field('kind', 'Kind of the type')>>
-    public function getKind(): __TypeKind;
+    public function getIntrospectionKind(): __TypeKind;
```
mwildehahn, line 13: nice, i like this for clarity

src/Types/NullableType.hack
```diff
@@ -0,0 +1,64 @@
...
+
+trait TNullableType implements INullableType {
+
+    /**
+     * GraphQL introspection doesn't use Nullable wrappers, so present this as an unwrapped type (delegate everything
+     * to the inner type).
+     */
+    <<__Override>>
+    final public function getIntrospectionKind(): Introspection\__TypeKind {
+        return $this->getInnerType()->getKind();
```
mwildehahn, line 18: we'd update these to `getIntrospectionKind`?

mwildehahn: Oh nvm. Read through the other one.

jjergus: No, this is intentional (this is the confusing part). When you introspect a type through a nullable wrapper, we return the *actual kind* of the inner type.

When you introspect the inner type directly, we pretend it's a non-null wrapper (getIntrospectionKind returns NON_NULL, getKind is not used at that point)

src/Types/Output/OutputType.hack
```diff
@@ -5,9 +5,23 @@ use namespace Slack\GraphQL;
 interface IOutputType {
     require extends BaseType;
 
-    public function unwrapType(): NamedOutputType;
+    public function unwrapType(): INamedOutputType;
 }
 
+interface IOutputTypeFor<TExpected, TResolved> extends IOutputType {
```
ianhoffman, line 11: What is the benefit of using this interface (and `IInputTypeFor`) as opposed to the traits directly? E.g., could the constructor on `FieldDefinition` just accept an object typed as `TOutputTypeFor<TRet, TResolved>`?

src/Types/Output/CompositeType.hack
```diff
@@ -0,0 +1,20 @@
+namespace Slack\GraphQL\Types;
+
+use namespace HH\Lib\Vec;
+use namespace Slack\GraphQL;
+
+abstract class CompositeType extends NamedType {
```
ianhoffman, line 6: +1

src/Types/Input/InputObjectType.hack
```diff
@@ -5,7 +5,8 @@ use namespace Slack\GraphQL;
 use namespace Graphpinator\Parser\Value;
 
 <<__ConsistentConstruct>>
-abstract class InputObjectType extends NamedInputType {
+abstract class InputObjectType extends NamedType {
+    use TNamedInputType;
```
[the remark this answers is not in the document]
jjergus: When we have a combination of interface + matching trait, we need something to distinguish the two, so I went with the I prefix for the interface and T for the trait. I can rename it to whatever you like (e.g. one of them could have no prefix).

src/Types/Output/OutputType.hack
jjergus: Traits can't be used as type annotations :( That's why we need a matching interface for every trait.

src/Types/Input/InputObjectType.hack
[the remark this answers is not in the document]
mwildehahn: I don't have a preference. Internally we suffix types with `_t`. When I was doing https://github.com/slackhq/hack-json-schema I had read through a couple of the open source libs from facebooked and learned ya'll prefix types with `T` and interfaces with `I` which I've liked.

src/Types/InputOutput/LeafType.hack
```diff
@@ -2,8 +2,10 @@ namespace Slack\GraphQL\Types;
 
 use namespace Slack\GraphQL;
 
-abstract class LeafOutputType extends NamedOutputType {
-    abstract const type TCoerced as mixed = this::THackType;
+abstract class LeafType extends NamedType {
+    use TNonNullableType;
+    use TNamedInputType;
+    use TNamedOutputType;
```
ianhoffman, line 8: It's pretty confusing that `coerce` is used to coerce from a Hack type to a GQL type, but `coerceValue` and `coerceNonVariableNode` are used to coerce from GQL types to Hack types. Can we maybe rename one side of these? Maybe `coerce` could be `serializeValue` or something like that?

src/Introspection/__Type.hack
ianhoffman: i think this is also needed more functionality too - for `TNonNullableType` these mostly return null, whereas for `TNullableType` they delegate to the inner type's `getName()`, `getDescription()`, `getOfType()`, etc, methods.

src/Types/Input/InputObjectType.hack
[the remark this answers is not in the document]
jjergus: It might be a bit confusing that we use the T prefix for both traits and for generic type params/type constants.

IIRC the T prefix is actually required by Hack for type constants and/or generic type params, but maybe that's changed or I remember wrong.

jjergus:
new commit renames and cleans up things

nullableO -> nullableOutput
nullableListOfO -> nullableOutputListOf
coerce -> serialize
TCoerced -> TResolved (consistent with parent classes/interfaces)
moved docblocks from traits to interfaces
removed final from methods in final classes
improve some docblocks

src/Types/InputOutput/Scalar/IntType.hack
```diff
@@ -38,4 +38,14 @@ final class IntInputType extends ScalarInputType {
...
         return $node->getRawValue();
     }
+
+    <<__Override>>
+    protected function serialize(int $value): int {
+        \Slack\GraphQL\assert(
+            $value >= self::MIN_SAFE_VALUE && $value <= self::MAX_SAFE_VALUE,
+            'Integers must be in 32-bit range, got %d',
+            $value,
+        );
```
ianhoffman, line 48: This assertion could be pulled into a private helper since it's used like 3 times in this class.

ianhoffman:
I like it!

ianhoffman (approved):

src/Introspection/__Type.hack
mwildehahn: YEa, I hadn't read that far 😄

src/Types/Output/OutputType.hack
ianhoffman: TIL

mwildehahn (approved):
Amazing job trying out these different directions. I think this ended up in a great spot!

mwildehahn:
@jjergus going to merge this so i can do some work off the latest changes

PR #86, opened 2021-05-21
Support `possibleTypes` introspection for interfaces

mwildehahn:
We need access to the schema to be able to resolve name -> `__Type` for these two introspection fields:

```
  # OBJECT only
  interfaces: [__Type!]

  # INTERFACE and UNION only
  possibleTypes: [__Type!]
```

I liked this approach more than the previous approach where we injected schema if a `<<GraphQL\Field(...)>>` accepted `BaseSchema` as an argument. The schema is an implementation detail we shouldn't be exposing to any `GraphQL\Field` just because we happen to use those to power introspection. 

Schema doesn't technically store any state but we had a mix of static and instance methods so I just decided to remove the static methods since I didn't think they were giving us a lot.

I broke this up into two commits since most of this PR is just threading `$schema` through everywhere. The second commit implements `possibleTypes` which was trivial.

### DIFF
--- a/src/Types/Output/InterfaceType.hack
+++ b/src/Types/Output/InterfaceType.hack
@@ -13,7 +13,6 @@ abstract class InterfaceType extends CompositeType {
 
     <<__Override>>
     final public function getPossibleTypes(): vec<GraphQL\Introspection\__Type> {
-        return Vec\map(static::POSSIBLE_TYPES, $type ==> $this->schema->getIntrospectionType($type))
-            |> Vec\filter_nulls($$);
+        return Vec\map(static::POSSIBLE_TYPES, $type ==> $this->schema->getIntrospectionType($type) as nonnull);
     }
 }
--- a/src/Types/Output/InterfaceType.hack
+++ b/src/Types/Output/InterfaceType.hack
@@ -4,9 +4,16 @@ use namespace HH\Lib\{Dict, Vec};
 use namespace Slack\GraphQL;
 
 abstract class InterfaceType extends CompositeType {
+    abstract const keyset<string> POSSIBLE_TYPES;
 
     <<__Override>>
     final public function getKind(): GraphQL\Introspection\__TypeKind {
         return GraphQL\Introspection\__TypeKind::INTERFACE;
+    }
+
+    <<__Override>>
+    final public function getPossibleTypes(): vec<GraphQL\Introspection\__Type> {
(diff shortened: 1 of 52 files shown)


src/Types/Output/InterfaceType.hack
```diff
@@ -4,9 +4,16 @@ use namespace HH\Lib\{Dict, Vec};
...
+    abstract const keyset<string> POSSIBLE_TYPES;
 
     <<__Override>>
     final public function getKind(): GraphQL\Introspection\__TypeKind {
         return GraphQL\Introspection\__TypeKind::INTERFACE;
     }
+
+    <<__Override>>
+    final public function getPossibleTypes(): vec<GraphQL\Introspection\__Type> {
+        return Vec\map(static::POSSIBLE_TYPES, $type ==> $this->schema->getIntrospectionType($type))
```
ianhoffman, line 16: Would it be better to cast the return value of `getIntrospectionType` to `as nonnull`?

tests/IntrospectionTest.hack
```diff
@@ -8,6 +8,47 @@ final class IntrospectionTest extends PlaygroundTest {
...
+                                'name' => 'ImplementInterfaceB',
+                            ],
+                            dict[
+                                'name' => 'ImplementInterfaceC',
+                            ],
+                        ],
+                    ],
+                ],
+            ),
+            'validate interface introspection' => tuple(
```
ianhoffman, line 33: Should this test have a different name from the above test?

ianhoffman (approved):

src/Types/Output/InterfaceType.hack
mwildehahn: Yea, I was trying to be safe but it is probably better to fail hard rather than silently failing here.

tests/IntrospectionTest.hack
mwildehahn: copy 🍝

mwildehahn:
@jjergus curious if you have thoughts on how we're passing down $schema here

jjergus:
I haven't really thought this through so there might be some reason why this wouldn't work, but couldn't we store a list of classnames in POSSIBLE_TYPES instead of GraphQL type names? It could give us better type-safety (we could type it as `keyset<classname<Types\ObjectType>>` to guarantee we don't accidentally put a scalar/enum/whatever there, or a type that doesn't exist) and we may not need the schema at all.

mwildehahn:
damn, great call! https://github.com/mwildehahn/hack-graphql/pull/88/files

PR #88, opened 2021-05-21
Possible types support without passing the schema

mwildehahn:


src/Types/Output/InterfaceType.hack
```diff
@@ -4,9 +4,15 @@ use namespace HH\Lib\{Dict, Vec};
...
+    abstract const keyset<classname<GraphQL\Types\ObjectType>> POSSIBLE_TYPES;
 
     <<__Override>>
     final public function getKind(): GraphQL\Introspection\__TypeKind {
         return GraphQL\Introspection\__TypeKind::INTERFACE;
     }
+
+    <<__Override>>
+    final public function getPossibleTypes(): vec<GraphQL\Introspection\__Type> {
+        return Vec\map(static::POSSIBLE_TYPES, $type ==> $type::nonNullable()->nullableForIntrospection());
```
mwildehahn, line 16: This wasn't super intuitive. At first I did `nonNullable` and that doesn't work because then we return these with the non null wrapper. `$type::nullableOutput` works but I think the intent is that we return `nullableForIntrospection`.

ianhoffman (commented):
Nice!

src/Codegen/Builders/InterfaceBuilder.hack
```diff
@@ -1,9 +1,12 @@
 namespace Slack\GraphQL\Codegen;
 
+use namespace HH\Lib\{Vec, Str};
```
ianhoffman, line 3: Unused

src/Types/Output/InterfaceType.hack
jjergus: It doesn't really matter, but I think `nullableOutput()` is probably preferable (less confusing).

I added `nullableForIntrospection()` only so that we had something to call from places that don't know if a given type is an input/output type, so they don't know whether to call `nulllableOutput()` or `nullableInput()`. There should be no need to call `nullableForIntrospection()` from anywhere except the few introspection-specific places  that currently do.

src/Types/Output/InterfaceType.hack
```diff
@@ -4,9 +4,15 @@ use namespace HH\Lib\{Dict, Vec};
...
 abstract class InterfaceType extends CompositeType {
+    abstract const keyset<classname<GraphQL\Types\ObjectType>> POSSIBLE_TYPES;
 
     <<__Override>>
     final public function getKind(): GraphQL\Introspection\__TypeKind {
         return GraphQL\Introspection\__TypeKind::INTERFACE;
     }
+
+    <<__Override>>
+    final public function getPossibleTypes(): vec<GraphQL\Introspection\__Type> {
```
jjergus, line 15: We could probably make this (and similar introspection methods) return `INullableType` to make it impossible to do the wrong thing.

Alternatively, maybe this could return `vec<ObjectType>` (for maximum type-safety + if this is reused from validation or elsewhere, that will probably be more useful). This is of course the wrong thing for introspection, but we can take care of that by doing a `Vec\map` inside `TNullableType::getIntrospectionPossibleTypes`, which has the bonus advantage that all "confusing introspection stuff" stays in one place -- we can't prevent TNullableType from being confusing but we can prevent everything else :)

jjergus (approved):

PR #89, opened 2021-05-21
implement field collection (aka defragmentation)

jjergus:
This makes it possible to execute queries with fragments, including all the fun cases around nested fragments and nested overlapping selection sets (see new test case).

One missing part is correctly codegenning the `INTERACES` const on object types (similarly to `POSSIBLE_TYPES` on interfaces) -- I assumed @mwildehahn is probably working on that already so for now I just hardcoded the value that makes the test pass :)

### DIFF
--- a/tests/FragmentTest.hack
+++ b/tests/FragmentTest.hack
@@ -1,0 +1,61 @@
+use function Facebook\FBExpect\expect;
+
+use namespace Slack\GraphQL;
+
+final class FragmentTest extends PlaygroundTest {
+
+    <<__Override>>
+    public static function getTestCases(): this::TTestCases {
+        return dict[
+            'various interesting cases' => tuple(
+                '
+                    query {
+                        one: user(id: 1) { ...BotFrag, ...UserFrag, team { id } }
+                        two: bot(id: 2)  { ...BotFrag, ...UserFrag, team { id } }
+                    }
+
+                    fragment UserFrag on User {
+                        team { name }
+                        ... on Human {
+                            favorite_color
+                            team {
+                                aliased_field: name
+                                description(short: true)
+                            }
+                        }
+                        ...BotFrag
+                    }
+
+                    fragment BotFrag on Bot {
+                        primary_function
+                        team {
+                            aliased_field: description(short: false)
+                            num_users
+                        }
+                    }
+                ',
+                dict[],
+                dict[
+                    'one' => dict[
+                        'team' => dict[
+                            'name' => 'Test Team 1',
+                            'aliased_field' => 'Test Team 1',
+                            'description' => 'Short description',
+                            'id' => 1,
+                        ],
+                        'favorite_color' => 'BLUE',
+                    ],
+                    'two' => dict[
+                        'primary_function' => 'send spam',
+                        'team' => dict[
+                            'aliased_field' => 'Much longer description',
+                            'num_users' => 3,
+                            'name' => 'Test Team 1',
+                            'id' => 1,
+                        ],
+                    ],
+                ],
+            ),
+        ];
+    }
+}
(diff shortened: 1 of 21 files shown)


mwildehahn:
😄 https://github.com/mwildehahn/hack-graphql/pull/90

src/Types/NonNullableType.hack
```diff
@@ -8,7 +8,7 @@ interface INonNullableType {
     public function getName(): ?string;
     public function getDescription(): ?string;
     public function getFields(bool $include_deprecated = false): ?vec<Introspection\__Field>;
-    public function getInterfaces(): ?vec<Introspection\__Type>;
+    public function getInterfaces(): ?dict<string, classname<InterfaceType>>;
```
mwildehahn, line 11: I just did this as a vec, I can update to a dict. I also like returning just the classname here and then having the `getIntrospectionInterfaces` deal with init'ing it 👍

tests/FragmentTest.hack
```diff
@@ -0,0 +1,61 @@
...
+    public static function getTestCases(): this::TTestCases {
+        return dict[
+            'various interesting cases' => tuple(
+                '
+                    query {
+                        one: user(id: 1) { ...BotFrag, ...UserFrag, team { id } }
+                        two: bot(id: 2)  { ...BotFrag, ...UserFrag, team { id } }
+                    }
+
+                    fragment UserFrag on User {
```
mwildehahn, line 17: ahhh i'm so excited to use this!

mwildehahn (approved):

src/Types/NonNullableType.hack
jjergus: I went with the dict so that

- we can check existence by GraphQL type name in O(1)
- classnames instead of instances for performance, so that we don't construct all the instances during validation/execution

mwildehahn:
Yea makes sense. Updated the other PR to match this

ianhoffman (approved):

jjergus:
I added one more commit which improves some names and adds more comments for parts that I thought were interesting.

Also fixes an inconsistency with the spec, where if there are issues that should have been caught by validation, they should be ignored during execution, not throw redundant errors.

jjergus:
> Just curious: what do we use these additional nodes for? Is there any reason to preserve the additional nodes instead of just building a map from field name (or alias) to one "representative" node?

Validation guarantees they have the same name and arguments, but they can have different selection sets.

```GraphQL
{
  a {
    subfield1
  }
  ...ExampleFragment
}

fragment ExampleFragment on Query {
  a {
    subfield2
  }
  b
}
```

The response here should be:

```JSON
"a": {
  "subfield1": "...",
  "subfield2": "..."
},
"b": "..."
```

We could merge the inner selection sets in FieldCollector (e.g. return a single field node for "a" with both subfields) but that would be more expensive, especially if we don't know if the inner selection set will be needed at all (e.g. "a" could resolve to `null`).

ianhoffman:
Got it. But the way the code is written, I don't see where we're handling the selection sets for nodes which aren't the first item in each list returned from `FieldCollector`? E.g., it seems like we're calling `C\firstx($grouped_field_nodes)` during resolution, not iterating over the grouping.

I'm sure I'm missing something.

ianhoffman:
(And when we're not calling `C\firstx`, we're just threading down the grouped nodes (which become the list of parents for the next resolver). So I still don't understand where the 2nd node in each grouping is handled.)

jjergus:
> And when we're not calling C\firstx, we're just threading down the grouped nodes (which become the list of parents for the next resolver)

Well you found the exact place :)

The "next resolver" passes the list of nodes to FieldCollector, and FieldCollector iterates over all of them: https://github.com/mwildehahn/hack-graphql/blob/main/src/FieldCollector.hack#L29

ianhoffman:
Aha! Ok, got it. Thanks 🙇‍♂️

PR #90, opened 2021-05-22
Support introspecting interfaces an object implements

mwildehahn:
Same flow as possible types on an interface. Biggest change was needing to swap to collect all interfaces before building objects.

### DIFF
--- a/src/Codegen/Builders/ObjectBuilder.hack
+++ b/src/Codegen/Builders/ObjectBuilder.hack
@@ -1,7 +1,13 @@
 namespace Slack\GraphQL\Codegen;
 
 use namespace HH\Lib\{Vec, Str};
-use type Facebook\HackCodegen\{CodegenClass, HackCodegenFactory, CodegenClassConstant, HackBuilderValues};
+use type Facebook\HackCodegen\{
+    CodegenClass,
+    HackCodegenFactory,
+    CodegenClassConstant,
+    HackBuilderValues,
+    HackBuilderKeys,
+};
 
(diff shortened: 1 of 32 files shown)


src/Codegen/Builders/ObjectBuilder.hack
```diff
@@ -19,6 +43,20 @@ final class ObjectBuilder<TField as IFieldBuilder> extends CompositeBuilder<TFie
...
             Vec\map_with_key($ts['fields'], ($name, $ts) ==> new ShapeFieldBuilder($name, $ts)),
+            $hack_class_to_graphql_interface,
         );
     }
+
+    private function generateInterfacesConstant(HackCodegenFactory $cg): CodegenClassConstant {
+        $interfaces = dict[];
+        foreach ($this->hack_class_to_graphql_interface as $hack_class => $graphql_type) {
+            if (\is_subclass_of($this->hack_type, $hack_class)) {
+                $interfaces[$graphql_type] = Str\format('%s::class', $hack_class);
```
jjergus, line 54: I think we want the generated GraphQL class name here (which AFAIK is always equivalent to `$graphql_type` right now), not the original Hack class, or not?

```suggestion
                $interfaces[$graphql_type] = Str\format('%s::class', $graphql_type);
```

jjergus (approved):

src/Codegen/Builders/ObjectBuilder.hack
mwildehahn: Good catch, updated with a test too to catch these being named differently

PR #91, opened 2021-05-22
Pagination

ianhoffman:
This is a pretty gnarly (IMO) first stab at pagination. You'll notice that I changed _a lot_ of things, especially when it comes to codegen. That's because connections and edges accept type parameters, which we've not handled yet during codegen. And to be honest, I wasn't sure _how_ to handle type parameters, so instead I just special-cased the connection and edge objects. What this means practically is that I've decoupled code generation from reflection. I _think_ this is a good thing. Basically, we no longer need reflection methods and classes to generate code — we can use them if we have them in hand, and we usually do, but for the case of generating classes which subclass classes with type arguments, or generating methods which return generics, it's better to just explicitly tell the code generator what we want to produce. As such:

* I created a new `FieldBuilder` base class which generates a GraphQL field from a spec.
* We still have MethodFieldBuilder and ShapeFieldBuilder. Before, these built GQL fields from Hack methods and Hack shape fields, respectively; now, they merely build GraphQL fields which resolve using Hack methods and Hack shape fields respectively. They can be built from any specification.
* We provide factory entrypoints for conveniently obtaining a MethodFieldBuilder or ShapeFieldBuilder from a Hack method or shape field. However, we don't _need_ to use these entrypoints — we can also tell the builders exactly what we want by calling their constructors directly. And this is what we do when generating connections and edges.
* I had to introduce a rather horrible special-cased `ConnectionFieldBuilder` which sets the pagination arguments on the connection immediately returned from calling the user-provided resolver. I'm not sure how else to do this short of making the user code accept the pagination arguments itself and pass them into the Connection constructor, which seems fine but also like the kind of thing the framework can automate.

I'm sort of conflicted about all these changes. On the one hand, they make it simpler to write user code, which we want. On the other hand, they introduce more complexity into codegen. I think separating out codegen from reflection is good and will create more flexibility down the line. I'm less sure about the special-cased handling of methods which return connections.

The reason for these changes, again, is that I just wasn't sure how else to handle codegen for connections. Since the user-defined connection subclass simply indicates by a type parameter what the edges returned from `getEdges` should be, the actual `getEdges` method has a generic return type. But we want to generate edge and connection types which are not generic, and I don't think we can use reflection to generate non-generic types from generic types. So I figured it was best to just tell the code generator exactly what we want for the special cases of connections and edges. Furthermore, I think connections and edges are the only objects in the GraphQL specification which require this special handling, so we won't need to do this all over the place 😓 .

I am, however, very open to alternative suggestions! I briefly explored using attributes like we do for introspection. This would make user code significantly more verbose (not just because of the additional attributes, but because we'd need to provide non-generic versions of all generic methods in each connection subclass). It might, however, eliminate most of the changes to codegen introduced by this PR. I figured the extra complexity in the framework is worth the enhanced developer experience, but could be persuaded otherwise.

### DIFF
--- a/src/Codegen/Builders/ObjectBuilder.hack
+++ b/src/Codegen/Builders/ObjectBuilder.hack
@@ -9,14 +9,14 @@ use type Facebook\HackCodegen\{
     HackBuilderKeys,
 };
 
-final class ObjectBuilder<TField as IFieldBuilder> extends CompositeBuilder<TField> {
+class ObjectBuilder extends CompositeBuilder {
     const classname<\Slack\GraphQL\Types\ObjectType> SUPERCLASS = \Slack\GraphQL\Types\ObjectType::class;
 
     <<__Override>>
     public function __construct(
         \Slack\GraphQL\__Private\CompositeType $type_info,
         string $hack_type,
-        vec<TField> $fields,
+        vec<FieldBuilder> $fields,
         private dict<string, string> $hack_class_to_graphql_interface,
     ) {
         parent::__construct($type_info, $hack_type, $fields);
@@ -26,25 +26,6 @@ final class ObjectBuilder<TField as IFieldBuilder> extends CompositeBuilder<TFie
     public function build(HackCodegenFactory $cg): CodegenClass {
         return parent::build($cg)
             ->addConstant($this->generateInterfacesConstant($cg));
-    }
-
-    public static function fromTypeAlias<T>(
-        \Slack\GraphQL\ObjectType $type_info,
-        \ReflectionTypeAlias $type_alias,
-        dict<string, string> $hack_class_to_graphql_interface,
-    ): ObjectBuilder<ShapeFieldBuilder<T>> {
-        $ts = $type_alias->getResolvedTypeStructure();
-        invariant(
-            $ts['kind'] === \HH\TypeStructureKind::OF_SHAPE && !idx($ts, 'nullable', false),
-            'Output objects can only be generated from type aliases of a non-nullable shape type, got %s.',
-            TypeStructureKind::getNames()[$ts['kind']],
-        );
-        return new ObjectBuilder(
-            $type_info,
-            $type_alias->getName(),
-            Vec\map_with_key($ts['fields'], ($name, $ts) ==> new ShapeFieldBuilder($name, $ts)),
-            $hack_class_to_graphql_interface,
-        );
     }
 
     private function generateInterfacesConstant(HackCodegenFactory $cg): CodegenClassConstant {
@@ -58,5 +39,78 @@ final class ObjectBuilder<TField as IFieldBuilder> extends CompositeBuilder<TFie
         return $cg->codegenClassConstant('INTERFACES')
             ->setType('dict<string, classname<Types\InterfaceType>>')
             ->setValue($interfaces, HackBuilderValues::dict(HackBuilderKeys::export(), HackBuilderValues::literal()));
+    }
+
+    public static function fromTypeAlias<T>(
+        \Slack\GraphQL\ObjectType $type_info,
+        \ReflectionTypeAlias $type_alias,
+    ): ObjectBuilder {
+        $ts = $type_alias->getResolvedTypeStructure();
+        invariant(
+            $ts['kind'] === \HH\TypeStructureKind::OF_SHAPE && !idx($ts, 'nullable', false),
+            'Output objects can only be generated from type aliases of a non-nullable shape type, got %s.',
+            TypeStructureKind::getNames()[$ts['kind']],
+        );
+        return new ObjectBuilder(
+            $type_info,
+            $type_alias->getName(),
+            Vec\map_with_key(
+                $ts['fields'],
+                ($name, $ts) ==> FieldBuilder::fromShapeField($name, $type_alias->getName(), $ts),
+            ),
+            dict[], // Objects generated from shapes cannot implement interfaces
+        );
+    }
+
+    public static function forConnection(string $name): ObjectBuilder {
+        return new ObjectBuilder(
+            new \Slack\GraphQL\ObjectType($name, $name), // TODO: Description
+            $name,  // hack type
+            vec[ // fields
+                new MethodFieldBuilder(shape(
+                    'name' => 'edges',
+                    'method_name' => 'getEdges',
+                    'output_type' => shape(
+                        'type' => 'UserEdge::nonNullable()->nullableOutputListOf()',
+                        'needs_await' => true,
+                    ),
+                    'declaring_type' => $name,
+                    'parameters' => vec[]
+                )),
+                new MethodFieldBuilder(shape(
+                    'name' => 'pageInfo',
+                    'method_name' => 'getPageInfo',
+                    'output_type' => shape('type' => 'PageInfo::nullableOutput()', 'needs_await' => true),
+                    'declaring_type' => $name,
+                    'parameters' => vec[]
+                )),
+            ],
+            dict[], // Connections do not implement any interfaces
+        );
+    }
+
+    public static function forEdge(string $hack_type): ObjectBuilder {
+        $name = $hack_type.'Edge';
+        return new ObjectBuilder(
+            new \Slack\GraphQL\ObjectType($name, $hack_type.' Edge'), // TODO: Description
+            'Slack\GraphQL\Pagination\Edge<\\'.$hack_type.'>',  // hack type
+            vec[ // fields
+                new MethodFieldBuilder(shape(
+                    'name' => 'node',
+                    'method_name' => 'getNode',
+                    'output_type' => shape('type' => $hack_type.'::nullableOutput()'),
+                    'declaring_type' => $name,
+                    'parameters' => vec[],
+                )),
+                new MethodFieldBuilder(shape(
+                    'name' => 'cursor',
+                    'method_name' => 'getCursor',
+                    'output_type' => shape('type' => 'Types\StringType::nullableOutput()'),
+                    'declaring_type' => $name,
+                    'parameters' => vec[],
+                )),
+            ],
+            dict[],
+        );
     }
 }
--- a/src/Codegen/Builders/ObjectBuilder.hack
+++ b/src/Codegen/Builders/ObjectBuilder.hack
@@ -75,20 +75,21 @@ class ObjectBuilder extends CompositeBuilder {
                         'needs_await' => true,
                     ),
                     'declaring_type' => $name,
-                    'parameters' => vec[]
+                    'parameters' => vec[],
                 )),
                 new MethodFieldBuilder(shape(
                     'name' => 'pageInfo',
                     'method_name' => 'getPageInfo',
                     'output_type' => shape('type' => 'PageInfo::nullableOutput()', 'needs_await' => true),
                     'declaring_type' => $name,
-                    'parameters' => vec[]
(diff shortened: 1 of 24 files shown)


src/Codegen/Generator.hack
```diff
@@ -229,7 +234,21 @@ final class Generator {
...
-            if (!C\is_empty($class->getAttributes())) {
+            if (Str\ends_with($class->getName(), 'Connection')) {
+                // TODO: Assert that any class which subclasses Connection<T> has a name ending in `Connection`.
+                invariant(
+                    is_connection_type(new \ReflectionClass($class->getName())),
+                    '%s must subclass "Connection"',
+                    $class->getName(),
+                );
+                $objects[] = ObjectBuilder::forConnection($class->getName());
+                $type_param = $class->getASTx()
```
ianhoffman, line 245: This is kind of gnarly and won't work on classes which subclass a non-generic connection. E.g., if `FriendConnection` subclasses `UserConnection`, and `UserConnection` extends `Connection<User>`, then we won't be able to deduce that `FriendConnection` extends `Connection<User>` as well.

Maybe we need to walk the inheritance tree? Ew.

src/Codegen/Types.hack
```diff
@@ -119,10 +119,16 @@ function get_output_class(string $hack_type): ?string {
     try {
         $rc = new \ReflectionClass($hack_type);
 
-        $graphql_object = $rc->getAttributeClass(\Slack\GraphQL\ObjectType::class) ??
-            $rc->getAttributeClass(\Slack\GraphQL\InterfaceType::class);
+        $graphql_object = $rc->getAttributeClass(\Slack\GraphQL\ObjectType::class);
         if ($graphql_object) {
             return $graphql_object->getType();
+        } elseif (is_connection_type($rc)) {
```
ianhoffman, line 125: Connection types aren't annotated with GraphQL attributes so we need to check the separately.

src/Pagination/Connection.hack
```diff
@@ -0,0 +1,196 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Constuct the correct edges and pageInfo objects from the result of calling the user-implemented `paginate` 
+     * method with the pagination args.
+     */
+    <<__Memoize>>
+    final private async function doPaginate(): Awaitable<shape('edges' => vec<Edge<T>>, 'pageInfo' => PageInfo)> {
```
ianhoffman, line 143: Very open to better names for this method.

src/Pagination/Connection.hack
```diff
@@ -0,0 +1,196 @@
...
+ */
+abstract class Connection<T> {
+
+    /**
+     * Paginate the data set per the pagination args, which contain cursors and limits.
+     *
+     * For example, if the pagination args contain a `before` cursor with value `foo` and a `last` limit with value 5, 
+     * `paginate` should return edges wrapping the 5 items immediately prior to the item with cursor `foo`.
+     */
+    abstract protected function paginate(PaginationArgs $args): Awaitable<vec<Edge<T>>>;
```
ianhoffman, line 56: I decided to diverge from my spec and make the implementation construct each edge, rather than having the framework do this for them. This is to allow edges which have cursors that are not derived from a node's attribute but from some other data point, like the node's index in a list.

tests/gen/Human.hack
```diff
@@ -34,6 +35,39 @@ final class Human extends \Slack\GraphQL\Types\ObjectType {
...
+              'type' => Types\IntType::nullableInput(),
+              'default_value' => null,
+            ),
+            'last' => shape(
+              'name' => 'last',
+              'type' => Types\IntType::nullableInput(),
+              'default_value' => null,
+            ),
+          ],
+          async ($parent, $args, $vars) ==> (await $parent->getFriends())->setPaginationArgs(
```
ianhoffman, line 64: This is the special cased resolver generated by the `ConnectionFieldBuilder` method. Yuck, but maybe it's OK? 🤔

src/playground/UserConnection.hack
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        $last = $args['last'] ?? null;
+        if ($last) {
+            $start_id = Math\maxva($start_id, $end_id - $last);
+        }
+
+        $edges = vec[];
+        for ($i = $start_id; $i < $end_id; $i++) {
+            $edges[] = new GraphQL\Pagination\Edge(
```
mwildehahn, line 32: This feels ok to me! It would be nice to just have this return `Awaitable<vec<User>>` but I think the ability to paginate a simple vec will be useful.

src/Codegen/Generator.hack
```diff
@@ -30,19 +30,29 @@ final class Generator {
...
+    private function __construct(private MultiParser $parser, private self::TGeneratorConfig $config) {
         $this->cg = new HackCodegenFactory($config['codegen_config'] ?? new HackCodegenConfig());
     }
 
+    private static async function init(
+        DefinitionFinder\BaseParser $parser,
+        self::TGeneratorConfig $config
+    ): Awaitable<Generator> {
+        $parsers = vec[$parser];
+        $parsers[] = await DefinitionFinder\TreeParser::fromPathAsync(__DIR__.'/../Introspection');
```
mwildehahn, line 42: Nice

src/Codegen/Builders/ObjectBuilder.hack
```diff
@@ -18,7 +18,61 @@ final class ObjectBuilder<TField as IFieldBuilder> extends CompositeBuilder<TFie
...
+    public static function forConnection(string $name): ObjectBuilder {
+        return new ObjectBuilder(
+            new \Slack\GraphQL\ObjectType($name, $name), // TODO: Description
+            $name,  // hack type
+            vec[ // fields
+                new MethodFieldBuilder(shape(
+                    'name' => 'edges',
+                    'method_name' => 'getEdges',
+                    'output_type' => shape(
+                        'type' => 'UserEdge::nonNullable()->nullableOutputListOf()',
```
mwildehahn, line 37: This should be dynamic?

src/Codegen/Builders/ObjectBuilder.hack
```diff
@@ -18,7 +18,61 @@ final class ObjectBuilder<TField as IFieldBuilder> extends CompositeBuilder<TFie
...
+                    'method_name' => 'getPageInfo',
+                    'output_type' => shape('type' => 'PageInfo::nullableOutput()', 'needs_await' => true),
+                    'declaring_type' => $name,
+                    'parameters' => vec[],
+                )),
+            ],
+        );
+    }
+
+    // TODO: It should be possible to create user-defined edges which contain additional fields.
```
mwildehahn, line 54: I was thinking about this, you can always just create the connections and edges yourself if you want this kind of customizability.

jjergus (approved):
> I'm sort of conflicted about all these changes. On the one hand, they make it simpler to write user code, which we want. On the other hand, they introduce more complexity into codegen.

I think you're optimizing for the right thing. Only a small number of people will ever work on the codegen, and there could potentially be a lot of users, and connections are very common in GraphQL.

Also I really appreciate all these refactors you're doing for the codegen, it's been so nice whenever I needed to make any changes recently, and it's only getting better! Your decision to add the intermediate representations for types/fields sounds exactly right (basically whenever you have X inputs and Y outputs, you want to find a way to structure the code such that the amount of code is O(X+Y), not O(X*Y) -- which this exactly achieves).

tests/gen/Human.hack
jjergus: Looks fine to me, why yuck? :)

You might want to add a test field that has extra arguments, so we will have `coerce` calls here both inside `getFriends` and inside `setPaginationArgs`.

src/Pagination/Connection.hack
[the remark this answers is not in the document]
ianhoffman: Yeah, good idea

src/Pagination/Connection.hack
```diff
@@ -0,0 +1,196 @@
...
+            $page = Vec\take($page, $first - 1);
+        }
+
+        $last = $this->args['last'] ?? null;
+        if ($last is nonnull) {
+            $page_info['hasPreviousPage'] = C\count($page) > $last - 1;
+            if (Shapes::keyExists($this->args, 'before')) {
+                $page_info['hasNextPage'] = await $this->hasPageAfterBeforeCursor($this->args['before']);
+            }
+            $page = Vec\drop($page, 1);
```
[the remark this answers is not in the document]
ianhoffman: Ah, good catch. We only want this if we returned `$last` items. Same for the above (when we check `$first`).

Will add more tests.

tests/gen/Human.hack
ianhoffman: Just feels yucky to have one special case among code that is generally not full of special cases.

That said, good idea on that test case. Will add.

src/Codegen/Generator.hack
ianhoffman: I like the idea of just using a type constant; making connection leaf classes generic over multiple types isn't something we need to (or even can, I think) support.

src/playground/UserConnection.hack
ianhoffman: yeah. let's keep thinking about this. It feels OK to me too, but would love to find a way to support both.

src/Codegen/Builders/ObjectBuilder.hack
ianhoffman: Doh

ianhoffman:
> I think you're optimizing for the right thing. Only a small number of people will ever work on the codegen, and there could potentially be a lot of users, and connections are very common in GraphQL.

Totally. It's just a very different mindset from being a product engineer, where you're always trying to keep product code fairly simple so that the next generation of engineers aren't too confused.
 
> Also I really appreciate all these refactors you're doing for the codegen, it's been so nice whenever I needed to make any changes recently, and it's only getting better! Your decision to add the intermediate representations for types/fields sounds exactly right (basically whenever you have X inputs and Y outputs, you want to find a way to structure the code such that the amount of code is O(X+Y), not O(X*Y) -- which this exactly achieves).

Thanks! I figure that the codegen is what we, the maintainers, have to live with and work in, so it's doing ourselves a favor to make it halfway decent.

mwildehahn (commented):
Amazing!!

mwildehahn (approved):

mwildehahn:
Going to merge this to avoid conflicts

src/Codegen/Builders/ObjectBuilder.hack
jjergus: Adding extra fields to the connection type and/or the edge type is probably worth supporting at some point (especially edge fields, e.g. the timestamp of when the edge was created is very common for all kinds of edges), but it can wait.

src/playground/UserConnection.hack
jjergus: We need the implementor to provide a node and and a cursor for each item (+any extra fields if we support them in the future) and this seems like a pretty natural way to do that (it might be a bit nicer if we could take care of automatically encoding the cursor though -- not just for convenience but mainly to prevent bugs from people forgetting that).

But if we see people writing too much boilerplate around this, a possible option would be:

1. subclass returns a list of generic type TItem instead of a list of Edge
2. subclass implements a method that transforms TItem -> Edge (presumably there would be some common ones already implemented, so most developers only need to do 1 -- otherwise this wouldn't save much)

PR #94, opened 2021-05-24
Port NoUndefinedVariablesRule

ianhoffman:
Ports NoUndefinedVariablesRule from the reference implementation (code [here](https://github.com/graphql/graphql-js/blob/main/src/validation/rules/NoUndefinedVariablesRule.js)). This means following references to named fragments to determine if they reference undefined variables.

I followed the reference implementation pretty much 1:1. To do this, I had to make a couple substantial changes:
* Add a unique id to every node. These IDs allow us to memoize nodes, which is useful since a lot of the work done in this PR will be used by other rules too, and we don't want to re-compute it each time.
* Handle fragments in the ASTVisitor.

### DIFF
--- a/src/Validation/ValidationContext.hack
+++ b/src/Validation/ValidationContext.hack
@@ -15,10 +15,6 @@ type VariableUsage = shape(
 
 final class ValidationContext {
     private vec<\Slack\GraphQL\UserFacingError> $errors = vec[];
-    private dict<int, vec<VariableUsage>> $variable_usages = dict[];
-    private dict<int, vec<VariableUsage>> $recursive_variable_usages = dict[];
-    private dict<int, vec<Parser\Fragment\Fragment>> $recursively_referenced_fragments = dict[];
-    private dict<int, vec<Parser\FragmentSpread\NamedFragmentSpread>> $named_fragment_spreads = dict[];
 
     public function __construct(
         private classname<\Slack\GraphQL\BaseSchema> $schema,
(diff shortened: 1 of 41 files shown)


src/Validation/ValidationContext.hack
```diff
@@ -1,15 +1,28 @@
...
+use namespace HH\Lib\{C, Str, Vec};
+use namespace Graphpinator\Parser;
+use namespace Slack\GraphQL\Types;
+use type HH\Lib\Ref;
+use type Slack\GraphQL\__Private\CallbackVisitor;
+use type Slack\GraphQL\__Private\TypeInfo;
+use type Slack\GraphQL\__Private\Utils\Stack;
+
+type VariableUsage = shape(
+    'node' => Parser\Value\VariableRef,
```
ianhoffman, line 12: This will eventually contain more info which is why it's a shape today.

jjergus:
> Add a unique id to every node. These IDs allow us to memoize nodes

This is great, should be useful for other validation rules too and possibly other things (e.g. if we wanted to memoize the results of FieldCollector which can sometimes be useful).

I think you could avoid most of the boilerplate by doing something like this:

```Hack
abstract class Node implements \HH\IMemoizeParam {

    private static int $nextID = 0;
    private int $id;

    public function __construct(
        private \Graphpinator\Common\Location $location
    ) {
      $this->id = self::$nextID;
      ++self::$nextID;
    }
```

This will also guarantee unique IDs even if we end up parsing multiple GraphQL queries in the same request/script, which could be important.

ianhoffman:
Good call @jjergus — that's much cleaner. Adopted.

ianhoffman:
That truly makes this PR so much easier to review.

jjergus:
For the implementation, I'm not sure if this makes sense (I haven't really thought it through), but it feels a bit confusing that most of the complex logic happens in `ValidationContext` and not the actual validation rule class.

It also seems that some of the functions (`getVariableUsages` etc.) are doing their own AST traversal, which should be the responsibility of the visitor (I mean, we have the visitor so that we wouldn't have to duplicate the same AST traversal logic in each validation rule class -- which could also increase the risk of bugs, e.g. what if in some specific validation rule we forget to traverse into directives, etc.)

Now I understand this rule can't be implemented in a completely straightforward way (whenever we enter a variable node, we immediately determine if it's good and output an error) since variables can occur in fragments referenced from other fragments or operations that are only visited later...but maybe something like this would work:

(everything happens in `NoUndefinedVariablesRule`)

1. enter operation: set $this->current_operation to its name, set $this->operation_variables[$name] = $op->getVariables()
2. leave operation: set $this->current_operation = null
3. enter fragment: set $this->current_fragment to its name
4. leave fragment: set $this->current_fragment = null
5. enter named fragment spread: store that $current_operation or $current_fragment depends on this $frag_spread->getName()
6. enter variable: store that $current_operation or $current_fragment depends on $var->getName()
7. leave document (i.e. do this once at the very end): use stored information to find all invalid variable references, output errors

I understand we may need to reuse some of this information for other validation rules (check fragment cycles, check variable usages match variable types...) so we may need to extract it into a shared place somewhere instead of `NoUndefinedVariablesRule`, but that shared place could also implement `ASTVisitor` and follow the logic above (take advantage of the traversal done by the visitor), e.g. this could be `FragmentInfo` (for 1--5) and `VariableInfo` (for 6), similarly to how we have `TypeInfo`.

tests/Validation/NoUndefinedVariablesRuleTest.hack
```diff
@@ -0,0 +1,343 @@
...
+                    shape(
+                        'message' => 'Variable "$b" is not defined by operation "Bar".',
+                        'location' => shape('line' => 11, 'column' => 36),
+                    ),
+                    shape(
+                        'message' => 'Variable "$c" is not defined by operation "Bar".',
+                        'location' => shape('line' => 14, 'column' => 29),
+                    ),
+                ],
+            ),
```
jjergus, line 340: Maybe add a test with a fragment loop.

```GraphQL
fragment Foo {
  ...Foo
}
```

There is another validation rule to detect these and output errors, but we should make sure it doesn't cause us to go into an infinite loop in this rule or elsewhere.

ianhoffman:
Hmm. These functions were ripped pretty much 1:1 from the reference PR. My thinking was:

1. I was being a bit lazy.
2. It'll be easier to continue porting things if we stick closely to the reference PR, especially as we onboard more people, because they'll just be able to follow the reference.

That said, I see what you're saying about just using the AST traversal instead of having "sub" traversals (even though the results of these sub-traversals are cached), and your proposal above makes sense broadly. I can take a stab at it and see how it shakes out.

jjergus (approved):
Makes sense, anyway feel free to address any of the comments later (or ignore them completely if you're busy/want to get other stuff done first), or address them here, whatever you prefer.

tests/Validation/NoUndefinedVariablesRuleTest.hack
ianhoffman: Added — it passes.

ianhoffman:
> Makes sense, anyway feel free to address any of the comments later (or ignore them completely if you're busy/want to get other stuff done first), or address them here, whatever you prefer.

Cool. I think I'm gonna merge this and we can always circle back and refactor later.

PR #95, opened 2021-05-25
Port NoUndefinedVariablesRule, version 2

ianhoffman:
@jjergus , this is a second attempt at porting the `NoUndefinedVariablesRule`, using new `FragmentInfo` and `VariableInfo` constructs per your feedback.

I'm honestly not sure which version I prefer. I do like that we lift all complex logic out of `ValidationContext` in this version. I don't love that we will now need to maintain these new classes instead of just copy+pasting the reference implementation, which I worry will make it harder to onboard new people (something i'm hoping to work on a bit in the coming weeks).

WDYT?


src/Validation/Validator.hack
```diff
@@ -1,30 +1,36 @@
...
+        $fragment_info = new FragmentInfo();
         $type_info = new TypeInfo($this->schema);
-        $ctx = new ValidationContext($this->schema, $type_info);
+        $variable_info = new VariableInfo($fragment_info);
+        $ctx = new ValidationContext($this->schema, $fragment_info, $type_info, $variable_info);
 
-        $visitor = new ParallelVisitor(Vec\map(static::$rules, $rule ==> new $rule($ctx)));
-        $visitor = $visitor->runAfter($type_info);
+        $visitor = new ParallelVisitor(Vec\concat(
+            vec[$fragment_info, $type_info, $variable_info],
```
ianhoffman, line 25: Don't love that we need to inject 3 "info" classes here

src/__Private/Visitor/FragmentInfo.hack
```diff
@@ -0,0 +1,59 @@
...
+        $fragments = dict[];
+        $stack = new Stack($this->operation_to_named_spread[$operation->getId()] ?? vec[]);
+        while ($stack->length() > 0) {
+            $fragment = $this->fragments[$stack->pop()->getName()] ?? null;
+            if ($fragment && !C\contains_key($fragments, $fragment->getName())) {
+                $fragments[$fragment->getName()] = $fragment;
+                foreach ($this->fragment_to_named_spread[$fragment->getId()] ?? vec[] as $frag) {
+                    $stack->push($frag);
+                }
+            }
```
ianhoffman, line 26: This loop is doing some repetitive work that could probably be memoized.

src/__Private/Visitor/VariableInfo.hack
```diff
@@ -0,0 +1,53 @@
...
+use namespace Graphpinator\Parser;
+
+final class VariableInfo extends ASTVisitor {
+    private ?Parser\Operation\Operation $current_operation = null;
+    private ?Parser\Fragment\Fragment $current_fragment = null;
+
+    private dict<int, vec<Parser\Value\VariableRef>> $operation_to_variables = dict[];
+    private dict<int, vec<Parser\Value\VariableRef>> $fragment_to_variables = dict[];
+
+    public function __construct(private FragmentInfo $fragment_info) {}
```
ianhoffman, line 12: Don't love that we need to inject FragmentInfo here. Maybe there's a better way

mwildehahn (approved):
This looks great. In general, I wouldn't worry about drifting from the reference implementation. I agree it's easier to bring on new devs, but the number of rules we have to implement is relatively stable and small. I also think we're going to want to hyper optimize for performance since validation is run on every request and imagine there will be places we'll want to refactor and not be bound by how the reference implementation implements something.

src/__Private/Visitor/FragmentInfo.hack
```diff
@@ -0,0 +1,59 @@
+namespace Slack\GraphQL\__Private;
+
+use namespace HH\Lib\C;
+use namespace Graphpinator\Parser;
+use type \Slack\GraphQL\__Private\Utils\Stack;
+
+final class FragmentInfo extends ASTVisitor {
```
ianhoffman, line 7: Couple thoughts @mwildehahn and @jjergus :

1. What do ya'll think of consolidating this and `VariableInfo` into one class, `DependencyTracker` or something like that?
2. Thought on throwing an error if someone calls `getFragments` (or any of these methods) before the AST has been fully traversed? Anything returned from these methods will be inaccurate till then.

jjergus:
> Thought on throwing an error if someone calls getFragments (or any of these methods) before the AST has been fully traversed? Anything returned from these methods will be inaccurate till then.

This is a very good point, probably worth addressing. Is there any case where VariableInfo or FragmentInfo is useful before we reach the end of the document? If not, maybe something like this would work:

1. Make VariableInfo and FragmentInfo unavailable from `enter` and `leave` (e.g. make them private in the ValidationContext)
2. Add a `finalize(VariableInfo, FragmentInfo)` method that validation rules can implement in addition to `enter` and `leave` (effectively equivalent to imlpementing `leave` for `ParsedRequest`, but allows us to provide this extra data here and not anywhere else)

jjergus:
> the number of rules we have to implement is relatively stable and small

Especially if you consider the number of rules that need variable/fragment dependency info (I think that's maybe 3 rules?)

src/__Private/Visitor/FragmentInfo.hack
jjergus: Any reason not to use recursion here instead of manually dealing with a stack? You could then put `<<__Memoize>>` on the recursive function (might have to be careful about infinite loops though).

src/__Private/Visitor/VariableInfo.hack
jjergus: You're right, these do seem pretty similar (e.g. both need to track current operation/fragment), merging them into DependencyInfo like you suggested seems reasonable.

jjergus (approved):

src/__Private/Visitor/FragmentInfo.hack
ianhoffman: > might have to be careful about infinite loops though

This was the reason. It's doable using an instance variable though, or by passing a set of visited frags through the call stack.

ianhoffman:
> 2\. Add a `finalize(VariableInfo, FragmentInfo)` method that validation rules can implement in addition to `enter` and `leave` (effectively equivalent to imlpementing `leave` for `ParsedRequest`, but allows us to provide this extra data here and not anywhere else)

Gonna do this.

mwildehahn (approved):

PR #97, opened 2021-05-25
Remaining introspection methods

mwildehahn:
This implements `enumValues` and `inputFields` which were the two big remaining introspection pieces remaining. There are still some TODOs scattered to support description, default value, etc.

I broke this up into two commits, first for enum values and next for input fields.

For input objects, we currently have a bug where you can't use a value with a generic parameter, ie this doesn't work: 

```
<<GraphQL\InputObjectType('IntrospectionRootInput', '')>>
type TRootInput = shape(
    'scalar' => string,
    'nested' => TNestedInput,
    'vec_of_nested' => vec<TNestedInput>,
);

<<GraphQL\InputObjectType('IntrospectionNestedInput', '')>>
type TNestedInput = shape(
    'string' => string,
    'vec_of_string' => vec<string>,
);
```

Because `THackType` is marked as `__Enforceable` so we can't use it with generics that are erased at runtime (ie `vec<TNestedInput>` and `vec<string>`. 

I was reading through this: https://docs.hhvm.com/hack/reified-generics/reified-generics to see if we could use reified generics but didn't see how to apply those to class constants.

### DIFF
--- a/src/Codegen/Builders/InputObjectBuilder.hack
+++ b/src/Codegen/Builders/InputObjectBuilder.hack
@@ -91,7 +91,7 @@ class InputObjectBuilder extends InputTypeBuilder<\Slack\GraphQL\InputObjectType
             );
 
             $name_literal = \var_export($field_name, true);
-            $type = input_type(($is_optional ? '?' : '').type_structure_to_type_alias($field_ts));
+            $type = input_type(type_structure_to_type_alias($field_ts));
 
             if ($is_optional) {
                 $hb->startIfBlock($get_if_condition($name_literal));
@@ -124,8 +124,7 @@ class InputObjectBuilder extends InputTypeBuilder<\Slack\GraphQL\InputObjectType
             $hb->addLine('return shape(')->indent();
             $hb->addLinef("'name' => %s,", \var_export($field_name, true));
 
-            $is_optional = $field_ts['optional_shape_field'] ?? false;
-            $type = input_type(($is_optional ? '?' : '').type_structure_to_type_alias($field_ts));
+            $type = input_type(type_structure_to_type_alias($field_ts));
             $hb->addLinef("'type' => %s,", $type);
             // TODO: description, defaultValue
             $hb->unindent()->addLine(');')->unindent();
--- a/src/Codegen/Builders/InputObjectBuilder.hack
+++ b/src/Codegen/Builders/InputObjectBuilder.hack
@@ -1,6 +1,6 @@
 namespace Slack\GraphQL\Codegen;
 
-use namespace HH\Lib\{Keyset, Str};
+use namespace HH\Lib\{Keyset, Str, Vec};
 use type Facebook\HackCodegen\{CodegenClass, CodegenMethod, HackBuilderValues, HackCodegenFactory};
 
 
@@ -65,6 +65,7 @@ class InputObjectBuilder extends InputTypeBuilder<\Slack\GraphQL\InputObjectType
                         $field_name ==> Str\format('assertValidVariableValue($fields[%s])', $field_name),
                     ),
                 ),
+            $this->generateGetInputValue($cg, $ts['fields']),
         ]);
     }
 
@@ -74,8 +75,8 @@ class InputObjectBuilder extends InputTypeBuilder<\Slack\GraphQL\InputObjectType
     private static function getFieldCoercionMethodBody<T>(
         HackCodegenFactory $cg,
         KeyedContainer<string, TypeStructure<T>> $fields,
-        (function (string): string) $get_if_condition,
-        (function (string): string) $get_method_call,
+        (function(string): string) $get_if_condition,
+        (function(string): string) $get_method_call,
     ): string {
         $hb = hb($cg)->addLine('$ret = shape();');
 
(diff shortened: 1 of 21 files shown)


src/Codegen/Builders/InputObjectBuilder.hack
```diff
@@ -44,21 +40,47 @@ class InputObjectBuilder extends InputTypeBuilder<\Slack\GraphQL\InputObjectType
             $name_literal = \var_export($field_name, true);
             $type = input_type(($is_optional ? '?' : '').type_structure_to_type_alias($field_ts));
 
+            $fields[$field_name] = shape(
```
mwildehahn, line 43: I was trying to keep this as a constant `FIELDS` but I think for this to work with list inputs, we'll want to do something similar to `getFieldDefinition` so we can take full advantage of the resolved `input_type` type here.

I think this would look something like:

```
function getInputFieldDefinition(string $name): GraphQL\Introspection\InputObjectType::TField {
  switch ($name) {
    case 'name':
      return shape('type' => Types\StringType::nonNullable());
    ...etc
   }
}
```

then in `__InputValue` we can just do:

```
public function getType(): __Type {
   return $this->field['type'];
}
```

and we don't have to leak `is_nullable` etc.

ianhoffman:
> For input objects, we currently have a bug where you can't use a value with a generic parameter, ie this doesn't work:

I ran into this when dealing with connections too. I couldn't figure out a way to make codegen work when dealing with methods that return generic types, or types parameterized by generic arguments. I actually think it's not a bug, and it's just something we shouldn't support: what would it mean to introspect a value of type `T`? What type of GraphQL object would `T` be? Basically, the GQL type system has no generics, so we probably shouldn't support generics either.

OTOH it would be really cool :)

mwildehahn:
I think this is less to do with representing a generic in graphql and more our ability to have an input object like: 

```
<<GraphQL\InputObjectType('IntrospectionNestedInput', '')>>
type TNestedInput = shape(
    'string' => string,
    'vec_of_string' => vec<string>,
);
```

This should translate to a graphql type:

```
type IntrospectionNestedInput {
  string: String
  vec_of_string: [String]
}
```

ianhoffman:
Oh got it. Yeah, it seems like we should be able to support that.

src/Codegen/Builders/InputObjectBuilder.hack
jjergus: Agreed, anything that has a type inside it probably can't be a constant.

src/Introspection/__InputValue.hack
```diff
@@ -3,17 +3,28 @@ namespace Slack\GraphQL\Introspection;
...
-    public function getDescription(): ?string;
+    public function getDescription(): ?string {
+        return $this->field['description'] ?? null;
+    }
 
     <<GraphQL\Field('type', 'Type of the field')>>
-    public function getType(): __Type;
+    public function getType(): __Type {
+        $type = $this->field['type'];
+        return $this->field['is_nullable'] ? $type::nullableInput() : $type::nonNullable();
```
jjergus, line 23: After the change discussed above, this check shouldn't be needed, since the method call chain returned by `input_type` should already correctly be `nullableInput`/`nonNullable`

src/Types/Input/InputObjectType.hack
```diff
@@ -11,7 +11,13 @@ abstract class InputObjectType extends NamedType {
 
     <<__Enforceable>>
     abstract const type THackType as shape(...);
-    abstract const keyset<string> FIELD_NAMES;
+    abstract const dict<string, self::TField> FIELDS;
+    const type TField = shape(
```
jjergus, line 15: Another minor advantage of the discussed change, we won't need this extra representation for the same thing (`__InputValue` class + this shape)

jjergus (approved):
> Because THackType is marked as __Enforceable so we can't use it with generics that are erased at runtime (ie vec<TNestedInput> and vec<string>.

Good catch, I missed this case when I marked it as `__Enforceable`. I'll fix it (will have to properly implement assertValidVariableValue).

src/Codegen/Types.hack
```diff
@@ -176,23 +175,26 @@ enum IO: string as string {
  * Get the type aias for the given type structure.
  */
 function type_structure_to_type_alias<T>(TypeStructure<T> $ts): string {
+    $prefix = ($ts['nullable'] ?? false) ? '?' : '';
```
mwildehahn, line 178: We need this to handle vecs of nullables

src/Codegen/Types.hack
```diff
@@ -176,23 +175,26 @@ enum IO: string as string {
...
         case TypeStructureKind::OF_BOOL:
-            return 'HH\bool';
+            return $prefix.'HH\bool';
         case TypeStructureKind::OF_VEC:
-            return Str\format('HH\vec<%s>', type_structure_to_type_alias($ts['generic_types'] as nonnull[0]));
+            return $prefix.Str\format('HH\vec<%s>', type_structure_to_type_alias($ts['generic_types'] as nonnull[0]));
         case TypeStructureKind::OF_ENUM:
-        //case TypeStructureKind::OF_UNRESOLVED: // not sure if this is needed
-            return $ts['classname'] as nonnull;
+        case TypeStructureKind::OF_INTERFACE:
```
mwildehahn, line 195: Not sure if we intentionally left this out, but needed this to resolve the `__Type` interface within the `__InputValue`

src/Codegen/Types.hack
jjergus: Just make sure we're still doing the right thing for regular (output) object types (I think this code is shared by both) -- but I don't see any changes to generated output objects so it's probably good.

src/Codegen/Types.hack
jjergus: I don't think anything is intentionally omitted here, we just didn't need it so far (also OF_CLASS probably, and OF_UNRESOLVED might be needed, I'm not sure when that applies -- probably when we have something that hasn't been loaded yet so it doesn't know if it's a class/interface/enum, not sure if that can happen).

jjergus (approved):

src/Codegen/Types.hack
mwildehahn: This was double wrapping in these cases: https://github.com/mwildehahn/hack-graphql/pull/97/commits/ebfdea44f9cd468b86322c8b0856128c56f02215 but `input_type` calls `unwrap_type` recursively so it handled it.

PR #98, opened 2021-05-25
Implement List Connection

ianhoffman:
Implement an out-of-the-box `ListConnection` which provides pagination over a static list. This is useful during development, as clients can test against a `ListConnection` until the actual connection is implemented.

### DIFF
--- a/src/Codegen/Builders/ObjectBuilder.hack
+++ b/src/Codegen/Builders/ObjectBuilder.hack
@@ -90,16 +90,16 @@ class ObjectBuilder extends CompositeBuilder {
     }
 
     // TODO: It should be possible to create user-defined edges which contain additional fields.
-    public static function forEdge(string $hack_type): ObjectBuilder {
-        $name = $hack_type.'Edge';
+    public static function forEdge(string $gql_type, string $hack_type, string $output_type): ObjectBuilder {
+        $name = $gql_type.'Edge';
         return new ObjectBuilder(
-            new \Slack\GraphQL\ObjectType($name, $hack_type.' Edge'), // TODO: Description
-            'Slack\GraphQL\Pagination\Edge<\\'.$hack_type.'>',  // hack type
+            new \Slack\GraphQL\ObjectType($name, $gql_type.' Edge'), // TODO: Description
+            'Slack\GraphQL\Pagination\Edge<'.$hack_type.'>',  // hack type
             vec[ // fields
                 new MethodFieldBuilder(shape(
                     'name' => 'node',
                     'method_name' => 'getNode',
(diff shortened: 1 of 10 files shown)


src/Codegen/Builders/ObjectBuilder.hack
```diff
@@ -90,16 +90,16 @@ class ObjectBuilder extends CompositeBuilder {
     }
 
     // TODO: It should be possible to create user-defined edges which contain additional fields.
-    public static function forEdge(string $hack_type): ObjectBuilder {
-        $name = $hack_type.'Edge';
+    public static function forEdge(string $gql_type, string $hack_type, string $output_type): ObjectBuilder {
```
ianhoffman, line 93: We now correctly handle Hack primitives.

src/playground/Playground.hack
```diff
@@ -204,4 +209,13 @@ abstract final class UserMutationAttributes {
             'team_id' => $team?->getId() ?? 1,
         ));
     }
+}
+
+
+final class AlphabetConnection extends GraphQL\Pagination\ListConnection {
```
mwildehahn, line 215: This is so good

mwildehahn (approved):

src/Codegen/Builders/ObjectBuilder.hack
jjergus: I can't imagine what that could be useful for :) but I agree it's probably better to not have special cases around which exact GraphQL types are allowed where.

src/Codegen/Types.hack
```diff
@@ -112,6 +100,22 @@ function get_input_class(string $hack_type): ?string {
     return null;
 }
 
+/**
+ * Get the GraphQL type to output for a hack type.
+ */
+function get_output_type(string $hack_type): ?string {
```
jjergus, line 106: Now that we don't have separate input/output classes, this switch statement (+enum type handling) could be shared between output_type and input_type

jjergus (approved):

src/Codegen/Builders/ObjectBuilder.hack
ianhoffman: I guess if you're paginating over a query that does something like `SELECT name...`. But yeah, not likely.

src/Codegen/Types.hack
ianhoffman: Good point!

src/Codegen/Builders/ObjectBuilder.hack
jjergus: That should probably be discouraged (the strength of GraphQL is that clients can query for exactly what data they need in a structured way, so returning only the name instead of a User object would violate that), but I guess it might be useful in some rare cases.

Issue #100, opened 2021-05-25
Support DataLoader pattern


ianhoffman:
We should support (and encourage) use of the DataLoader pattern to make querying efficient. We'll want to support this on a framework level because so that it integrates with the type-system. In short, I think we'll want something like this:

* Add a new abstract type, `GraphQL\Types\Promise` (or a better name) which expects as its Hack type a `Promise` parameterized by a Hack object.
* Fields annotated with `GraphQL\Field` should be allowed to return a `Promise<HackType>`. A `Promise<HackType>` can be obtained by, for example, calling `DataLoader::load($id)`.
* The `GraphQL\Types\Promise` class's `resolveAsync` invokes the resolver, returning the (unresolved) promise.

We can then do the following:
1. Resolve the request. We'll have a dict containing some number of un-resolved promises.
1. Batch load all pending data.
1. Walk the dict and resolve each promise. (We can avoid walking the entire dict each time.)
1. Repeat 1-3 until there are no more promises to resolve.

We'll need to make this approach work with introspection, too.

mwildehahn:
I don't know that this needs to be a part of this framework. I think we can implement something like the `DataLoader` pattern by leveraging async/await.

I looked into this once before and forgot some of the details, but I think we can do something like this:

```
$result = await DataLoader::getId(1)
```

Where the data loader can call:

```
		await Asio\later();
```

to allow other IDs to be collected before executing.

ianhoffman:
Oh interesting. That makes my work on #101 moot — I had thought we'd need to support something like this ourselves.

If `Asio\later` will work, that'd be great.

ianhoffman:
It would be good to understand exactly when `Asio\later()` kicks in. It's not like we're telling Hack to wake up the coroutine when a certain event occurs, so how does the event loop know when to resume these calls?

mwildehahn:
Found more docs on `Asio\later`: https://docs.hhvm.com/hack/asynchronous-operations/examples#batching

mwildehahn:
Although, this just shows an example, doesn't explain how it works

ianhoffman:
Seems like Jan had a good answer on Slack:

> I don't understand all the technical details, but my rough understanding is that this will put the current Awaitable at the end of the queue, so it will resume after every other pending Awaitable has had a chance to run (e.g. every other Awaitable has added some IDs to fetch to the current batch and called Asio\later).

Sounds like it's exactly what we want.

PR #101, opened 2021-05-26
Promises

ianhoffman:
This is an attempt to support promises, which we'll need if we want to use the DataLoader pattern to batch queries.

* Methods annotated with GraphQL\Field may return a Hack object with a valid GQL output representation wrapped in a `GraphQL\Promise<T>`; e.g., `Promise<Team>`.
* Codegen will special case these fields and see that they use a new `PromiseType<T>` instead of `T` directly.
* `PromiseType<T>` delegates all introspection to `T`.
* When resolved, `PromiseType<T>` returns a result of type `DeferredFieldResult<T>`.
* If an `ObjectType`, `ListOutputType`, or `NullableOutputType` detects that it will be returning a result which is or contains a `DeferredFieldResult`, it will return a `DeferredFieldResult` containing a callback which will continue resolution.
* The resolver checks if the result from calling `await $schema::resolveQuery` (or mutation) is itself a `DeferredFieldResult`. If so, it continues calling `resolveAsync()` on that field result until it receives a non-deferred result. Note that this will only happen once the top-level deferred result contains no more deferred results (because of the prior bullet point).
* This means that each time the resolver calls `resolveAsync()` on the deferred result, _all_ currently known deferred results will resolve. The next set of deferred results, should any exist, will be any new deferred result returned by calling `resolveAsync()` on each `DeferredFieldResult`. This means that between calls to `resolveAsync()` at the top-level is the perfect place to invoke the data loading framework and load batched data for all pending promises.

...Anyway. This is pretty rough, and I'm eager for feedback. Curious whether using `FieldResult` in this way makes sense or if there's a better approach. This is just what I came up with that seemed to result in the smallest amount of change to existing code.

### DIFF
--- a/src/Resolver.hack
+++ b/src/Resolver.hack
@@ -113,7 +113,7 @@ final class Resolver {
                 throw new \Error('Unsupported operation: '.$operation_type);
         }
 
-        while ($result->isDeferred()) {
+        while ($result is DeferredFieldResult<_>) {
             // This is where we'd batch load data for all pending promises
             $result = await $result->resolveAsync();
         }
--- a/src/Resolver.hack
+++ b/src/Resolver.hack
@@ -113,7 +113,16 @@ final class Resolver {
                 throw new \Error('Unsupported operation: '.$operation_type);
         }
 
-        return tuple($result->getValue(), $result->getErrors());
+        while ($result->isDeferred()) {
+            // This is where we'd batch load data for all pending promises
+            $result = await $result->resolveAsync();
+        }
+
+        if ($result is ValidFieldResult<_>) {
+            return tuple($result->getValue(), $result->getErrors());
+        } else {
+            return tuple(dict[], $result->getErrors());
+        }
     }
 
     private function coerceVariables(
--- a/src/Types/Output/ListOutputType.hack
+++ b/src/Types/Output/ListOutputType.hack
@@ -1,6 +1,6 @@
 namespace Slack\GraphQL\Types;
 
-use namespace HH\Lib\Vec;
+use namespace HH\Lib\{C, Dict};
 use namespace Slack\GraphQL;
 
 final class ListOutputType<TInner, TResolved>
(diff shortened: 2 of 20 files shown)


src/Resolver.hack
```diff
@@ -113,7 +113,16 @@ final class Resolver {
...
 
-        return tuple($result->getValue(), $result->getErrors());
+        while ($result is DeferredFieldResult<_>) {
+            // This is where we'd batch load data for all pending promises
+            $result = await $result->resolveAsync();
+        }
+
+        if ($result is ValidFieldResult<_>) {
+            return tuple($result->getValue(), $result->getErrors());
+        } else {
```
ianhoffman, line 123: I unfortunately lost some type safety when I changed the return type of `NullableOutputType::resolveAsync` from `ValidFieldResult<?T>` to `FieldResult<?T>`. However, now that `NullableOutputType::resolveAsync` can return a deferred result, I wasn't sure how to get around this.

src/Types/Output/ListOutputType.hack
```diff
@@ -23,18 +23,26 @@ final class ListOutputType<TInner, TResolved>
         vec<\Graphpinator\Parser\Field\IHasSelectionSet> $parent_nodes,
         GraphQL\ExecutionContext $context,
     ): Awaitable<GraphQL\FieldResult<vec<mixed>>> {
-        $ret = vec[];
-        $errors = vec[];
-        $is_valid = true;
-
-        $results = await Vec\map_async(
+        $results = await Dict\map_async(
```
ianhoffman, line 26: Had to change the result type here to a dict from indices to values so that we can merge it with deferred results later.

ianhoffman:
Per this, this is quite possibly unnecessary 😓 

https://github.com/mwildehahn/hack-graphql/issues/100#issuecomment-848483498

Well, was an interesting experiment.

ianhoffman:
Declining since we can just use `Asio\later()`.

PR #102, opened 2021-05-26
Add a utility for dumping the schema to JSON

mwildehahn:
This will let us read the JSON and dump the SDL using something like this: https://github.com/prisma-labs/get-graphql-schema/blob/master/src/index.ts#L75.


src/FieldDefinition.hack
```diff
@@ -1,13 +1,12 @@
 namespace Slack\GraphQL;
 
-use namespace HH\Lib\C;
+use namespace HH\Lib\{C, Vec};
 
 type ArgumentDefinition = shape(
```
mwildehahn, line 5: This is mostly the same as `Introspection\__InputValue` except in `Introspection\__InputValue` `default_value` is camel case and is a string value. This duplication seems worth it to keep `Introspection\__InputValue` a shape

jjergus: We can probably get rid of this and use your version everywhere. Validation only cares about the existence of a default value so it doesn't care if it's string or mixed.

src/BaseSchema.hack
```diff
@@ -53,5 +53,9 @@ abstract class BaseSchema implements Introspection\__Schema {
         return $type is nonnull ? $type::nonNullable()->nullableForIntrospection() : null;
     }
 
-    // TODO add method to create singleton
+    <<__Override>>
+    final public function getTypes(): vec<Introspection\__Type> {
+        // TODO: should we filter out private types? ie. anything with __
```
jjergus, line 58: Yes, the spec says those should not show up in introspection. The spec also requires that the `__` prefix is not used for anything else (although we're not enforcing it right now), so this check should be safe.

src/FieldDefinition.hack
```diff
@@ -70,4 +69,41 @@ final class FieldDefinition<TParent, TRet, TResolved> implements IResolvableFiel
...
+                    'name' => $arg['name'],
+                    'type' => $arg['type'],
+                );
+
+                if (Shapes::keyExists($arg, 'description')) {
+                    $input_value['description'] = $arg['description'];
+                }
+
+                if (Shapes::keyExists($arg, 'default_value')) {
+                    $input_value['defaultValue'] = (string)$arg['default_value'];
```
jjergus, line 102: This string needs to use GraphQL syntax (what you would write in a GraphQL query), e.g. string literals are quoted `"foo"`, lists are `[1, 2, 3]`, etc. I'm not sure if we have any default values where this matters, so it can possibly be addressed later.

A nice way to implement this would be to add a new method like `valueToString` to `IInputType`, then you can do `$arg['type']->valueToString($arg['default_value'])` here (or do that in the codegen and put the returned string in the generated files)

src/Introspection/Utilities.hack
```diff
@@ -0,0 +1,151 @@
...
+
+        /**
+         * Whether target GraphQL server support deprecation of input values.
+         * Default: false
+         */
+        ?'input_value_deprecation' => bool,
+    );
+
+    public static function getIntrospectionQuery(self::TIntrospectionOptions $options = shape()): string {
+        $descriptions = ($options['descriptions'] ?? false) ? 'description' : '';
```
jjergus, line 37: These would be nicer as actual GraphQL variables in the query, e.g.

```GraphQL
query IntrospectionQuery($include_descriptions: Boolean = true) {
  # ...
  description @include(if: $include_descriptions)
```

but we haven't implemented `@inculde` and `@skip` yet so we may have to address this later.

jjergus (approved):

src/FieldDefinition.hack
mwildehahn: Ah good call. Going to do this in a separate PR: https://github.com/mwildehahn/hack-graphql/issues/103

src/Codegen/Builders/Fields/FieldBuilder.hack
```diff
@@ -135,7 +135,7 @@ abstract class FieldBuilder {
 
                 $default_value = $param['default_value'] ?? null;
                 if ($default_value is nonnull) {
-                    $hb->addLinef("'default_value' => %s,", $default_value);
+                    $hb->addLinef("'defaultValue' => %s,", \var_export($default_value, true));
```
mwildehahn, line 138: Doing this as as top gap until: https://github.com/mwildehahn/hack-graphql/issues/103

PR #105, opened 2021-05-27
Support introspection from any schema

mwildehahn:
We were still hard coding the schema for introspection. The refactoring from @ianhoffman made this really simple 🙌 .


src/Codegen/Generator.hack
```diff
@@ -302,31 +305,23 @@ final class Generator {
...
+            $objects[] = new ObjectBuilder(
+                new \Slack\GraphQL\__Private\CompositeType('Mutation', 'Mutation'),
+                'Slack\\GraphQL\\Root',
+                vec(Dict\sort_by_key($mutation_fields)),
+                $hack_class_to_graphql_interface,
+            );
         }
 
-        return Vec\sort_by($objects, $object ==> $object->getGraphQLType());
+        return Vec\concat($objects, $inputs) |> Vec\sort_by($$, $object ==> $object->getGraphQLType());
```
ianhoffman, line 324: Why do we need a separate `inputs` vec? Can we just stick them in `$objects` too?

ianhoffman (approved):
nice

src/Codegen/Generator.hack
mwildehahn: Yea, this looks like it was just left over from something. Updated!

src/Codegen/Builders/Fields/IntrospectionSchemaFieldBuilder.hack
```diff
@@ -0,0 +1,27 @@
...
+    const type TField = shape(
+        'name' => string,
+        'output_type' => shape('type' => string, ?'needs_await' => bool),
+        ...
+    );
+
+    public function __construct() {
+        parent::__construct(shape(
+            'name' => '__schema',
+            'output_type' => shape('type' => '__Schema::nullableOutput()'),
```
ianhoffman, line 14: It would be nice at some point not to need hardcoded strings here, in case we ever want to change these method names.

src/Codegen/Generator.hack
```diff
@@ -302,28 +304,20 @@ final class Generator {
...
-            if (!C\is_empty($mutation_fields)) {
-                $top_level_objects[] = new ObjectBuilder(
-                    new \Slack\GraphQL\__Private\CompositeType('Mutation', 'Mutation'),
-                    'Slack\\GraphQL\\Root',
-                    vec(Dict\sort_by_key($mutation_fields)),
-                    $hack_class_to_graphql_interface,
-                );
-            }
+        $objects[] = new ObjectBuilder(
+            new \Slack\GraphQL\__Private\CompositeType('Query', 'Query'),
```
ianhoffman, line 308: Shouldn't this just be `new Slack\GraphQL\ObjectType`?

src/Codegen/Builders/Fields/IntrospectionSchemaFieldBuilder.hack
mwildehahn: This is a little annoying because `__Schema` is something that gets generated by introspection so we can't use `output_type` which uses reflection

ianhoffman: Maybe we could refactor `output_type` to not rely on introspection, just as we did for FieldBuilder?

Anyway, not a priority.

mwildehahn: Maybe! This seems like such a specific thing though that I'd rather do this for now 😄

PR #108, opened 2021-05-28
Fix possible types

### LINKED ISSUE #107: Fix `POSSIBLE_TYPES` when generating types from a namespace
The following code:

```hack
namespace ABC;

<<\Slack\GraphQL\InterfaceType('InterfaceFoo', 'Interface Foo')>>
abstract class InterfaceFoo {
	<<\Slack\GraphQL\Field('foo', 'foo')>>
	public function foo(): string {
		return 'foo';
	}
}

<<\Slack\GraphQL\ObjectType('ObjectBar', 'Object Bar')>>
final class ObjectBar extends InterfaceFoo {
	<<\Slack\GraphQL\Field('bar', 'bar')>>
	public function bar(): string {
		return 'bar';
	}
}
```

Produces the following incorrect codegen:

```hack

final class InterfaceFoo extends \Slack\GraphQL\Types\InterfaceType {

  const NAME = 'InterfaceFoo';
  const type THackType = \ABC\InterfaceFoo;
  const keyset<string> FIELD_NAMES = keyset[
    'foo',
  ];
  const keyset<classname<Types\ObjectType>> POSSIBLE_TYPES = keyset[
    ABC\ObjectBar::class,
  ];
  // Rest omitted for brevity
}
```

We should strip the namespace name from the class name when generating POSSIBLE_TYPES.

ianhoffman:
@mwildehahn , we should probably get this in the release we're about to ship internally.

Closes #107

### DIFF
--- a/src/Codegen/Builders/InterfaceBuilder.hack
+++ b/src/Codegen/Builders/InterfaceBuilder.hack
@@ -61,9 +61,9 @@ final class InterfaceBuilder extends CompositeBuilder {
 
     private function generatePossibleTypesConstant(HackCodegenFactory $cg): CodegenClassConstant {
         $possible_types = keyset[];
-        foreach ($this->hack_class_to_graphql_object as $hack_class => $_) {
+        foreach ($this->hack_class_to_graphql_object as $hack_class => $graphql_class) {
             if (\is_subclass_of($hack_class, $this->hack_type)) {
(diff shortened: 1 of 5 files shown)


mwildehahn (approved):
Ah, good catch

mwildehahn:
Will bump the version again and get this in

PR #109, opened 2021-05-28
Update introspection schema

ianhoffman:
Ugh, sorry I missed this @mwildehahn . We should update CI to fail if there's a diff after running tests.


mwildehahn:
Yea, I need to make this test better. Not a big deal for now.

PR #111, opened 2021-06-01
Handle fragments in FieldsOnCorrectTypeRule

ianhoffman:


tests/Validation/FieldsOnCorrectTypeRuleTest.hack
```diff
@@ -0,0 +1,131 @@
+use namespace Slack\GraphQL\Validation;
```
ianhoffman, line 1: Renamed and added a few more tests.

src/Introspection/__Type.hack
```diff
@@ -20,7 +20,7 @@ interface __Type {
     public function getIntrospectionDescription(): ?string;
 
     <<GraphQL\Field('fields', 'Fields of the type, only applies to OBJECT and INTERFACE')>>
-    public function getIntrospectionFields(bool $include_deprecated = false): ?vec<__Field>;
+    public function getIntrospectionFields(bool $includeDeprecated = false): ?vec<__Field>;
```
ianhoffman, line 23: `schema.json` won't generate without this change. (Or we'd need to pass in `include_deprecated` from the client, which seems wrong).

In general it might be nice to support auto-converting from snake case to camel case when converting to and from GraphQL arguments and Hack arguments?

src/__Private/Visitor/TypeInfo.hack
```diff
@@ -91,6 +91,16 @@ final class TypeInfo extends ASTVisitor {
                     throw new \Slack\GraphQL\UserFacingError("Unrecognized type: %s", $node->getType());
             }
             $this->type_stack->push($type);
+        } else if (
+            $node is Parser\FragmentSpread\InlineFragmentSpread ||
+            $node is Parser\Fragment\Fragment
+        ) {
+            $schema = $this->schema;
```
ianhoffman, line 98: We should pass down an instance of the schema, rather than the schema type itself, but I want to make that change separately.

mwildehahn (approved):

src/Introspection/__Type.hack
jjergus: It's probably easier/less confusing to adopt the same conventions for GraphQL as what you're using for Hack. There are only a very small number of these names that we can't change because they're required by the GraphQL spec.

ianhoffman: Makes sense!

PR #112, opened 2021-06-05
Support verbose field resolver errors

ianhoffman:
Adds a new option to the resolver, `verbose_errors`. When enabled, we also surface the cause of each `FieldResolverError`.

### DIFF
--- a/src/UserFacingError.hack
+++ b/src/UserFacingError.hack
@@ -2,11 +2,7 @@ namespace Slack\GraphQL;
 
 use namespace HH\Lib\{C, Str, Vec};
 
-function assert(
-    bool $condition,
-    Str\SprintfFormatString $message,
-    mixed ...$args
-): void {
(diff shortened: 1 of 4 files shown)


mwildehahn (approved):

src/UserFacingError.hack
```diff
@@ -20,22 +16,22 @@ class UserFacingError extends \Exception {
         'message' => string,
         ?'location' => shape('line' => int, 'column' => int),
         ?'path' => vec<arraykey>,
+        ?'cause' => shape(
```
jjergus, line 19: It doesn't really matter much, but you could put this inside `extensions` to more closely follow the spec: https://spec.graphql.org/draft/#sel-GAPHRPZCAACCBx6b

jjergus (approved):

src/UserFacingError.hack
ianhoffman: Nice, updated.

PR #115, opened 2021-07-26
Add request/response interfaces

ianhoffman:
Per discussion with @sarahhenkens, we want to create a convenient way to access the operation name for each GraphQL request (and other useful metadata, eventually). As such, we decided it would be convenient to create request/response interfaces exposing this information. So, instead of the following:

```hack
return await $resolver->resolve($input, $variables);
```

We would now do:

```hack
$request = GraphQL\Request::build($input, $variables);
$response = await $resolver->resolve($request);
return $response->toShape();
```

It's a bit more verbose, but somewhat more flexible, since we can now do things like `$request->getOperationName()` to retrieve the name of the operation which will be executed (if any).

We could provide a simpler interface:

```hack
$response = await $resolver->resolve($input, $variables);

// Get the operation name:
$operation_name = $response->getRequest()->getOperationName();

return $response->toShape();
```

However, I like having access to the `Request` object before calling `$resolver->resolve`, so that we can augment logs with the operation name before resolving the request. Definitely open to feedback here.

### DIFF
--- a/src/Request.hack
+++ b/src/Request.hack
@@ -1,0 +1,137 @@
+namespace Slack\GraphQL;
+
+use namespace Slack\GraphQL;
+use namespace HH\Lib\C;
+
+/**
+ * IRequest represents a request to the GraphQL server. The request may be wellformed or malformed;
+ * in the case of a malformed request, parsing has failed and the request will not be further validated
+ * or executed.
+ *
+ * This interface mostly exists as a way to pass useful metadata back to the client; e.g. callers can
+ * call `$request->getOperationName()` to discover the name of the GraphQL query or mutation which the
+ * request would execute.
+ */
+interface IRequest {
+    public function getVariables(): dict<string, mixed>;
+    public function getOperationName(): ?string;
+}
+
+abstract class Request implements IRequest {
+    /**
+     * Build a GraphQL request.
+     *
+     * A request contains at leat one operation. An operation name must be provided if the request
+     * contains multiple operations.
+     *
+     * @see https://spec.graphql.org/draft/#sec-Execution
+     */
+    final public static function build(
+        string $input,
+        dict<string, mixed> $variables = dict[],
+        ?string $operation_name = null,
+    ): IRequest {
+        try {
+            $query = self::parseQuery($input);
+            $operation = self::parseOperation($query, $operation_name);
+        } catch (UserFacingError $error) {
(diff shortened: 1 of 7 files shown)


src/Codegen/Builders/InputTypeBuilder.hack
```diff
@@ -1,5 +1,5 @@
 namespace Slack\GraphQL\Codegen;
```
ianhoffman, line 1: Some of the changes here are because we had never run `hackfmt` on the whole codebase before.

src/Request.hack
```diff
@@ -0,0 +1,137 @@
...
+            if ($location) {
+                $user_facing_error->setLocation($location);
+            }
+            $path = $e->getPath()?->jsonSerialize();
+            if ($path) {
+                $user_facing_error->setPath($path);
+            }
+            throw $user_facing_error;
+        }
+    }
```
ianhoffman, line 82: These are moved basically unaltered from the resolver.

src/Resolver.hack
```diff
@@ -5,104 +5,46 @@ use namespace HH\Lib\{C, Vec};
...
-            $request = $this->parseRequest($input);
-        } catch (GraphQL\UserFacingError $e) {
-            return shape('errors' => $this->formatErrors(vec[$e]));
+    public async function resolve(IRequest $request): Awaitable<IResponse> {
+        $response = new Response($request);
+        // If the request is malformed, we can't proceed with execution.
+        if ($request is MalformedRequest) {
+            return $response->withErrors($request->getErrors());
         }
+        $request as WellformedRequest;
```
ianhoffman, line 19: Is there a nicer way to do this?

jjergus: Not that I know of :(

This kind of conditional branching can be avoided by putting the logic in the request class (so we have `$request->resolve()` instead of `Resolver::resolve($request)` but that's not always practical (also the "visitor" pattern addresses this more generically but that would be an even bigger overkill here probably).

jjergus (approved):

src/Resolver.hack
ianhoffman: We'd need to update `resolve` to accept all the parameters one could use to instantiate a new resolver, in that case. I think I'll keep as-is for now and we can revisit. It would be nice to be able to write code like the following:

```hack
$request = GraphQL\Request::build($input, $variables);
$response = await $request->resolve();
```

mwildehahn (approved):

PR #123, opened 2021-09-28
User-defined directives / decorators

ianhoffman:
This PR is a prototype for how we might support user-defined directives / decorators which can be used to, for example, implement permissions. Specifically:

* Users can define custom field-level directives
* Users can annotate GraphQL fields with these directives

I still need to implement the following:
* Object-level directives
* Introspection for custom directives

Stuff I like:
* I mean the idea in general is just cool.

Stuff I don't like:
* AST parsing to figure out which directives are user-defined.
* Using `var_export` to instantiate the directives with the appropriate args within each GQL type.
* The fact that VSCode's "jump to" doesn't work with attributes.


src/Codegen/DirectivesFinder.hack
```diff
@@ -0,0 +1,118 @@
...
+                            |> Str\replace($$, 'vec ', 'vec')
+                            |> Str\replace($$, 'dict ', 'dict')
+                            |> Str\replace($$, 'shape ', 'shape')
+                    )
+                        |> Str\join($$, ', '),
+                );
+            }
+        }
+
+        return $directives;
```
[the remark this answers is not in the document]
admdikramr: I recommend always generating some of the code as part of the PR so that people can see and review the generated code as if it were human-written. Generally speaking it's best to hold generated code to the same high standards as human-written code.

src/Codegen/Builders/Fields/FieldBuilder.hack
```diff
@@ -14,6 +14,7 @@ abstract class FieldBuilder {
     abstract const type TField as shape(
         'name' => string,
         'output_type' => shape('type' => string, ?'needs_await' => bool),
+        'directives' => vec<string>,
```
admdikramr, line 17: In the spec directives can take arguments so you might want to future proof this and make it dict<string, mixed>. For sure you also don't want to double-up on the directives. And you'll also want key-based lookup instead of C\contains($vec, $thing) which is O(n) linear search.

src/Codegen/DirectivesFinder.hack
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        $directives = vec[];
+
+        // Get all use namespaces and types present
+        $file_path = $method->getFileName();
+        $use_decls = await self::getUseDeclarations($file_path);
+
+        // For each attribute, see if it matches a user-defined directive
+        // This requires building the qualified name, which is not
```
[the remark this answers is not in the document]
ianhoffman: Yeah, good call.

src/Codegen/DirectivesFinder.hack
```diff
@@ -0,0 +1,118 @@
...
+    )>> {
+        $ast = await \Facebook\HHAST\from_file_async(\Facebook\HHAST\File::fromPath($file_path));
+        return $ast->getDescendantsByType<\Facebook\HHAST\INamespaceUseDeclaration>()
+            |> Vec\map($$, $decl ==> {
+                $kind = $decl->getKind();
+                if ($kind is null || $kind->getText() === 'function') {
+                    return null;
+                }
+                $leaves = $decl->getClausesx()->toVec();
+                if ($decl is \Facebook\HHAST\NamespaceGroupUseDeclaration) {
```
admdikramr, line 94: HackAst is really slow and changes rapidly along with the language so a lot of it is dangerous to use. Can you use ReflectionClass::getNamespaceName() for this?

admdikramr: At FB we ban the use of HackAst for anything that's production-critical, including build systems.

ianhoffman: Agreed using HHAST is a bummer. I can play around with ReflectionClass more, but I have to admit I'm not very optimistic about it working.

If developers manually registered their custom directives, then we could use `Reflection{Class, Method}::getAttributeClass`, which actually does what we want. So maybe that's the interim approach until we support Facts: when creating a new directive, developers register it with the generator:

```hack
await GraphQL\Codegen\Generator::forPath(
    __DIR__.'/Fixtures',
    shape(
        'output_directory' => __DIR__.'/gen',
        'namespace' => 'Slack\GraphQL\Test\Generated',
        'custom_directives' => shape(
            'fields' => vec[Directives\HasRole::class],
            'objects' => vec[Directives\HasRole::class],
        ),
    ),
);
```

That's starting to feel like the cleanest way of doing this.

mwildehahn: Registering feels totally reasonable

ianhoffman: Ok, did that in the latest commit: https://github.com/mwildehahn/hack-graphql/pull/123/commits/355b1412fe3c4b358cc7a5d41d7b0c9f2aeffac8

I like where this is headed now. Will work on object directives and introspection next.

src/Codegen/DirectivesFinder.hack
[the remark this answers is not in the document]
admdikramr: No, I meant that I could not identify by looking at your link that it was a link to lower in this PR. :)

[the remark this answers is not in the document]
ianhoffman: Gotcha. I'd recommend reviewing this whole file: https://github.com/mwildehahn/hack-graphql/pull/123/files#diff-bc6416c0810e5e09cdcf817fbe7fc23de76a1884523575597eb7a6b4bc6b9861

It contains generated code for object and field-level directives.

PR #127, opened 2021-11-18
Fix bug for classes with non-graphql attribute classes

claycoleman:
I have a class with a non-graphql attribute class (`<<__ConsistentConstruct>>`) that was being picked up at this line and leading to codegen failures. the code expects my non-graphql class to exist in `$class_fields`, and this PR simply continues if the class has attribute classes but no graphql attribute classes. another solution could be to more narrowly tailor the directories that my codegen script checks, but seems like we should be able to handle this case fairly well.

### DIFF
--- a/src/Codegen/Generator.hack
+++ b/src/Codegen/Generator.hack
@@ -249,13 +249,9 @@ final class Generator {
                     $class->getName(),
                 );
                 $objects = Vec\concat($objects, $this->getConnectionObjects($class, $class_fields[$class->getName()]));
-            } elseif (!C\is_empty($class->getAttributes())) {
-                $fields = idx($class_fields, $class->getName());
-                if ($fields == null) {
-                  continue;
-                }
-                
+            } elseif (C\contains_key($class_fields, $class->getName())) {
                 $rc = new \ReflectionClass($class->getName());
+                $fields = $class_fields[$class->getName()];
                 $graphql_interface = $rc->getAttributeClass(\Slack\GraphQL\InterfaceType::class);
                 $graphql_object = $rc->getAttributeClass(\Slack\GraphQL\ObjectType::class);
                 invariant(
--- a/src/Codegen/Generator.hack
+++ b/src/Codegen/Generator.hack
@@ -250,8 +250,12 @@ final class Generator {
                 );
                 $objects = Vec\concat($objects, $this->getConnectionObjects($class, $class_fields[$class->getName()]));
             } elseif (!C\is_empty($class->getAttributes())) {
+                $fields = idx($class_fields, $class->getName());
+                if ($fields == null) {
+                  continue;
+                }
+                
                 $rc = new \ReflectionClass($class->getName());
-                $fields = $class_fields[$class->getName()];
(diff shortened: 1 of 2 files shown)


src/Codegen/Generator.hack
```diff
@@ -250,8 +250,12 @@ final class Generator {
                 );
                 $objects = Vec\concat($objects, $this->getConnectionObjects($class, $class_fields[$class->getName()]));
             } elseif (!C\is_empty($class->getAttributes())) {
+                $fields = idx($class_fields, $class->getName());
```
ianhoffman, line 253: A better fix might be 

```hack
} elseif (C\contains_key($class_fields, $class->getName()) {
```

on Line 252.

Mind making that change and then I'll merge this?

ianhoffman:
Nice find @claycoleman , I'm surprised we haven't run into this before. Have a small suggestion regarding your fix but the overall idea seems right.

src/Codegen/Generator.hack
ianhoffman: Would also be lovely to get a test for this, if you have the time. Just adding a class with a  `<<__ConsistentConstruct>>` annotation (and no other class or method-level annotations) to the test fixtures in `test/Fixtures/Playground.hack`, and ensuring nothing breaks, should be good.

claycoleman:
updated! Ran the tests without my fix to verify the error occurs, and sure enough: 

```
> vendor/bin/hacktest tests/


Summary: 0 test(s), 0 passed, 0 failed, 0 skipped, 0 error(s).

Fatal error: Uncaught exception 'OutOfBoundsException' with message 'Out of bounds dict access: invalid index "NonGraphQLClass"' in /Users/claycoleman/Developer/projects/pr-hack-graphql/src/Codegen/Generator.hack:254
Stack trace:
#0 /Users/claycoleman/Developer/projects/pr-hack-graphql/src/Codegen/Generator.hack(63): Slack\GraphQL\Codegen\Generator->collectObjects()
#1 /Users/claycoleman/Developer/projects/pr-hack-graphql/src/Codegen/Generator.hack(53): Slack\GraphQL\Codegen\Generator->generate()
#2 /Users/claycoleman/Developer/projects/pr-hack-graphql/tests/FixtureTest.hack(27): Slack\GraphQL\Codegen\Generator::forPath()
#3 /Users/claycoleman/Developer/projects/pr-hack-graphql/tests/FixtureTest.hack(28): FixtureTest::runCodegenAsync$memoize_impl()
#4 /Users/claycoleman/Developer/projects/pr-hack-graphql/tests/FixtureTest.hack(16): FixtureTest::runCodegenAsync()
#5 /Users/claycoleman/Developer/projects/pr-hack-graphql/vendor/hhvm/hacktest/src/Framework/HackTest.hack(61): FixtureTest::beforeFirstTestA%          
```

ran it with the changes:

```
................................................................................................................................................................................................................................................

Summary: 240 test(s), 240 passed, 0 failed, 0 skipped, 0 error(s).
```

ianhoffman (approved):

PR #128, opened 2022-01-06
Fix a minor bug where hasPrevious/NextPage returns null

brennemankyle:
Fix a minor bug where hasPreviousPage would return null, instead of false

### DIFF
--- a/src/Pagination/Connection.hack
+++ b/src/Pagination/Connection.hack
@@ -159,7 +159,10 @@ abstract class Connection {
     )> {
         $page = await $this->fetch($this->args);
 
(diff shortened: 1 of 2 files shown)


ianhoffman (approved):
LGTM, thanks!

ianhoffman:
Oh no, looks like CI is broken. Fred described the fix here: https://hacklang.slack.com/archives/CAAGK9GKW/p1641319124093000

> fwiw, my plan for the vendor/bin/ + composer changes it to change hh_server to always ignore vendor/bin, and cherry-pick to all supported releases

So I'll force merge for now

PR #129, opened 2022-01-11
Sort desc bug

brennemankyle:
When retrieving a list of items the project will grab 1 extra to cleverly figure out if there is a next page. However the project always assumes the list is sort ascending, when sending a descending list it doesn't remove the extra item but the item at the other end of the list.

To avoid adding sort direction to the project I removed the cleverness, and instead just fetch a single item of the next page to see if there is more

I also renamed
hasPageBeforeAfterCursor > hasNextPage
hasPageAfterBeforeCursor > hasPreviousPage
because I thought they were more clear


ianhoffman (approved):
Nice!

PR #130, opened 2022-01-12
Upgrade to HHVM 4.128

ianhoffman:


src/Types/NullableType.hack
```diff
@@ -40,13 +40,27 @@ trait TNullableType implements INullableType {
     <<__Override>>
     final public function getIntrospectionInterfaces(): ?vec<Introspection\__Type> {
```
ianhoffman, line 41: Had to change this because of https://github.com/facebook/hhvm/commit/362d29d0807169d4cf41487e070995d30f562314

PR #132, opened 2022-02-10
Page cursor

brennemankyle:
passes more params to hasNextPage/hasPreviousPage which allows for more complex overriding. This is especally necessary when using SQL with graphql. As you have to artificially change the sort order in SQL to retrieve things, thus changing which start or end cursor to use for the next/prev page

### DIFF
--- a/src/Pagination/Connection.hack
+++ b/src/Pagination/Connection.hack
@@ -89,7 +89,27 @@ abstract class Connection {
      * Whether a page exists before the `after` cursor, was such a cursor provided.
      * Can be overridden for efficiency.
      */
-    public async function hasNextPage(PaginationArgs $args, string $start_cursor, string $end_cursor): Awaitable<bool> {
+    protected async function hasNextPage(
+        PaginationArgs $args,
+        string $start_cursor,
+        string $end_cursor,
+    ): Awaitable<bool> {
+        if (
+            Shapes::keyExists($args, 'last') &&
+            !Shapes::keyExists($args, 'before') &&
+            !Shapes::keyExists($args, 'after')
+        ) {
+            return false;
+        }
+
+        if (
+            Shapes::keyExists($args, 'after') &&
+            !Shapes::keyExists($args, 'first') &&
+            !Shapes::keyExists($args, 'before')
+        ) {
+            return false;
+        }
+
         return C\count(await $this->fetch(shape('first' => 1, 'after' => $end_cursor))) > 0;
     }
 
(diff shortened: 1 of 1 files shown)


src/Pagination/Connection.hack
```diff
@@ -167,8 +170,16 @@ abstract class Connection {
...
             $page_info['startCursor'] = C\firstx($edges)->getCursor();
             $page_info['endCursor'] = C\lastx($edges)->getCursor();
-            $page_info['hasNextPage'] = await $this->hasNextPage($this->decodeCursor($page_info['endCursor']));
-            $page_info['hasPreviousPage'] = await $this->hasPreviousPage($this->decodeCursor($page_info['startCursor']));
+            $page_info['hasNextPage'] = await $this->hasNextPage(
+                $this->args,
+                $this->decodeCursor($page_info['startCursor']),
+                $this->decodeCursor($page_info['endCursor']),
+            );
+            $page_info['hasPreviousPage'] = await $this->hasPreviousPage(
```
ianhoffman, line 178: if no start cursor was provided, can we skip calling this function since it'll do unnecessary work?

src/Pagination/Connection.hack
```diff
@@ -90,17 +89,21 @@ abstract class Connection {
      * Whether a page exists before the `after` cursor, was such a cursor provided.
      * Can be overridden for efficiency.
      */
-    public async function hasNextPage(string $after_cursor): Awaitable<bool> {
-		return C\count(await $this->fetch(shape('first' => 1, 'after' => $after_cursor))) > 0;
-	}
+    public async function hasNextPage(PaginationArgs $args, string $start_cursor, string $end_cursor): Awaitable<bool> {
```
ianhoffman, line 92: IMO these function should be protected (and always should have been, my bad).

src/Pagination/Connection.hack
```diff
@@ -90,17 +89,21 @@ abstract class Connection {
...
 
     /**
      * Whether a page exists after the `before` cursor, was such a cursor provided.
      * Can be overridden for efficiency.
      */
-	public async function hasPreviousPage(string $before_cursor): Awaitable<bool> {
-		return C\count(await $this->fetch(shape('last' => 1, 'before' => $before_cursor))) > 0;
-	}
+    public async function hasPreviousPage(
+        PaginationArgs $args,
```
ianhoffman, line 101: I wonder if instead of passing PaginationArgs we can just access `$this->args`?

src/Pagination/Connection.hack
```diff
@@ -90,17 +89,49 @@ abstract class Connection {
...
+        string $start_cursor,
+        string $end_cursor,
+    ): Awaitable<bool> {
+        if (Shapes::keyExists($args, 'last') && !Shapes::keyExists($args, 'before')) {
+            return false;
+        }
+
+        if (
+            Shapes::keyExists($args, 'after') &&
```
ianhoffman, line 102: Why does it matter if `after` is provided here? Isn't it the same if we start at the first cursor?

ianhoffman (approved):

PR #133, opened 2022-02-25
allow better subclassing of Connection

brennemankyle:
allow subclassing Connection without having to instantiate abstract constants in an abstract subclass
aka Connection > SqlConnection > UserConnection (don't require SqlConnection to have to instantiate TNode)

Also make TNode abstract type enforceable

### DIFF
--- a/src/Codegen/Generator.hack
+++ b/src/Codegen/Generator.hack
@@ -242,7 +242,11 @@ final class Generator {
         }
 
         foreach ($classish_objects as $class) {
-            if (\is_subclass_of($class->getName(), \Slack\GraphQL\Pagination\Connection::class)) {
+            $rc = new \ReflectionClass($class->getName());
+            if (
+                \is_subclass_of($class->getName(), \Slack\GraphQL\Pagination\Connection::class) &&
+                $rc->getAttributeClass(\Slack\GraphQL\ObjectType::class)
+            ) {
                 invariant(
                     Str\ends_with($class->getName(), 'Connection'),
                     "All connection types must have names ending with `Connection`. `%s` does not.",
--- a/src/Pagination/Connection.hack
+++ b/src/Pagination/Connection.hack
@@ -50,6 +50,7 @@ type PaginationArgs = shape(
  * @see https://relay.dev/graphql/connections.htm for more information about GraphQL pagination.
  * @see src/playground/UserConnection.hack for an example.
  */
+ <<GraphQL\ObjectType('Connection', 'Connection')>>
 abstract class Connection {
 
     /**
@@ -57,6 +58,7 @@ abstract class Connection {
      *
      * This should be the Hack class over which you want to paginate.
      */
+     <<__Enforceable>>
     abstract const type TNode;
 
     /**
--- a/src/Pagination/Connection.hack
+++ b/src/Pagination/Connection.hack
@@ -50,7 +50,6 @@ type PaginationArgs = shape(
  * @see https://relay.dev/graphql/connections.htm for more information about GraphQL pagination.
  * @see src/playground/UserConnection.hack for an example.
  */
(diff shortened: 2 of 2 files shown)


src/Pagination/Connection.hack
```diff
@@ -50,13 +50,15 @@ type PaginationArgs = shape(
...
  */
+ <<GraphQL\ObjectType('Connection', 'Connection')>>
 abstract class Connection {
 
     /**
      * Node type for this connection.
      *
      * This should be the Hack class over which you want to paginate.
      */
+     <<__Enforceable>>
```
ianhoffman, line 61: Why does this need to be enforceable?

src/Codegen/Generator.hack
```diff
@@ -242,7 +242,11 @@ final class Generator {
         }
 
         foreach ($classish_objects as $class) {
-            if (\is_subclass_of($class->getName(), \Slack\GraphQL\Pagination\Connection::class)) {
+            $rc = new \ReflectionClass($class->getName());
+            if (
+                \is_subclass_of($class->getName(), \Slack\GraphQL\Pagination\Connection::class) &&
+                $rc->getAttributeClass(\Slack\GraphQL\ObjectType::class)
```
ianhoffman, line 248: Nice.

Should we get a test that this works? E.g., add a `FooConnection` abstract class subclassing `Connection`, and a `BarConnection` concrete class subclassing `FooConnection` and annotated with `ObjectType`?

src/Pagination/Connection.hack
```diff
@@ -50,13 +50,15 @@ type PaginationArgs = shape(
  * @see https://relay.dev/graphql/connections.htm for more information about GraphQL pagination.
  * @see src/playground/UserConnection.hack for an example.
  */
+ <<GraphQL\ObjectType('Connection', 'Connection')>>
```
ianhoffman, line 53: I actually realize this isn't necessary. We don't generate a GQL object for `Connection` so we don't need to tag the class with this attribute.

src/Pagination/Connection.hack
brennemankyle: if it's not enforceable I cannot do this in SqlConnection class `$node as this::TNode`
https://slack-github.com/slack/webapp/pull/128988/files#diff-eeef89f255be1dec860c0b1f779fe2eba85adfebf64bfe7e7c70ab8fef4a8703R47

ianhoffman: I still don’t follow. Looking at that code you shared, `$node as this::TNode` will explode whenever `this::TNode != this::TSqlNode` and is a no-op whenever `this::Node == this::TSqlNode`. No?

src/Codegen/Generator.hack
brennemankyle: hmm yeah not really sure where/ how to add this to a test? I added a fixture with sublcassing, but what file do I test it in?

ianhoffman: Just testing that the codegen looks correct should be fine!

brennemankyle: hmm how do I do that? sorry not familiar with out this sort of testing works

ianhoffman: Basically:
* Add the appropriate connection classes to `tests/Fixtures`.
* Run tests using `make test`. This will do the codegen.
* Look at the generated files and confirm they look correct. E.g., if you declared an abstract `FooConnection` and a concrete `BarConnection`, we should end up with one generated class in a `BarConnection.hack`.
Also happy to pair!

ianhoffman: You also have failing tests... we'll need to annotate `UserConnection.hack` with the `ObjectType` attribute, and ideally look at that attribute in `Generator::getConnectionObjects`. Again, happy to pair.

